PR #1, opened 2020-08-28
bring nonspatial environment and timeline back

eagmon:
This PR brings the ```nonspatial_environment``` and ```timeline``` processes back into ```composition.py```. Pytests are not all passing because of a registry exception that we've been seeing independently in ```vivarium-cell```: 

<pre>Exception: registry already contains an entry for timeline: <class 'vivarium.processes.timeline.TimelineProcess'>></pre>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ with open("README.md", 'r') as readme:
 
 setup(
     name='vivarium-core',
-    version='0.0.6',
+    version='0.0.8',
     packages=[
         'vivarium',
         'vivarium.core',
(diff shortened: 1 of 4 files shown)


prismofeverything:
Okay I'll take a look and find a better solution for the registry. Thanks @eagmon!

prismofeverything:
Okay worked this out. It turns out pytest requires that each module have an __init__.py so it doesn't get confused as to what the hierarchy is. I just released `vivarium-core==0.0.10` with all pytests passing. Still rethinking the registry.... the main issue is populating it through top level code, not the registry in general. We may just need to populate it by direct function call somewhere rather than in top level code, though that makes it hard for tests to find it (each test would then have to make that call). Let's go over possibilities again in another meeting.

PR #8, opened 2020-09-11
[WIP] bug with node=None in schema_topology

eagmon:
In running some chemotaxis experiments, I came across a bug in which the environment attempts to update an agent that has already been deleted.  The ```schema_topology``` in ```Store``` uses ```node = self.get_path(path)``` with this agent's path, and retrieves a ```None```. As a partial fix, I have it skip the update if the node is ```None```. Though with this fix, the environment still continues to try updating the agent for the remained of the experiment. Maybe we should check if a node has been marked as deleted and update the environment's subschema?

### DIFF
--- a/vivarium/core/experiment.py
+++ b/vivarium/core/experiment.py
@@ -44,9 +44,15 @@ from vivarium.core.registry import (
 
 INFINITY = float('inf')
 VERBOSE = False
+EMPTY_UPDATES = None, None, None
 
 log.basicConfig(level=os.environ.get("LOGLEVEL", log.WARNING))
 
+
+def starts_with(l, sub):
+    return len(sub) <= len(l) and all(
+        l[i] == el
+        for i, el in enumerate(sub))
 
 # Store
 def key_for_value(d, looking):
@@ -95,6 +101,35 @@ def dissoc(d, removing):
         key: value
         for key, value in d.items()
         if key not in removing}
+
+
+def delete_in(d, path):
+    if len(path) > 0:
+        head = path[0]
+        if len(path) == 1:
+            # at the node to be deleted
+            del d[head]
+        elif head in d:
+            down = d[head]
+            delete_in(d[head], path[1:])
+
+
+def depth(tree, path=()):
+    '''
+    Create a mapping of every path in the tree to the node living at
+    that path in the tree.
+    '''
+
+    base = {}
+
+    for key, inner in tree.items():
+        down = tuple(path + (key,))
+        if isinstance(inner, dict):
+            base.update(depth(inner, down))
+        else:
+            base[down] = inner
+
+    return base
 
 
 def without(d, removing):
@@ -167,7 +202,6 @@ class Store(object):
         self.divider = None
         self.emit = False
         self.sources = {}
-        self.deleted = False
         self.leaf = False
         self.serializer = None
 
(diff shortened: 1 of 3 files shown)


vivarium/core/experiment.py
```diff
@@ -418,16 +418,17 @@ def get_path(self, path):
         '''
 
         if path:
-            step = path[0]
-            if step == '..':
+            first_step = path[0]
```
prismofeverything, line 421: I think `step` is still the appropriate name here, since the function is recursive this represents each step along the path eventually.

vivarium/core/experiment.py
```diff
@@ -805,7 +806,11 @@ def schema_topology(self, schema, topology):
                     if path is None:
                         path = (key,)
                     node = self.get_path(path)
-                    state[key] = node.schema_topology(subschema, {})
+                    if node:
+                        state[key] = node.schema_topology(subschema, {})
```
prismofeverything, line 810: I think we still want to set the state key here since it was requested in the topology even though the value is `None`, in order to mirror the structure of the topology passed in. `None` is a legitimate value.

prismofeverything (commented):
Hey @eagmon, I'm not sure the problem is here, looks like a process has not received the information that an agent was removed and is trying to update it? Either way the `None` value is legitimate and needs to be handled on the process side, unless I'm missing something here.

prismofeverything:
Or perhaps the problem is this element is not being removed from the topology upon deletion? Do you have a test case for this I can use?

vivarium/core/experiment.py
eagmon: ```step``` was shadowing the ```ipdb``` function called ```step```.

vivarium/core/experiment.py
eagmon: In the case I found ```node``` was equal to ```None```, which can't have ```schema_topology``` called on it.

eagmon:
@prismofeverything -- I have a test case in ```vivarium-chemotaxis```. Yeah, the element was not being removed.

vivarium/core/experiment.py
prismofeverything: You can always access shadowed ipdb methods using !, like `!step`.

prismofeverything:
Okay, I just made a bunch of improvements in this branch to track deletions to topology, as well as tracking added and deleted processes. This simplifies some logic in update (where it was walking the state tree to find processes and derivers), so should be a performance improvement as well. 

I also solved the mystery of why this wasn't triggering before.... the `transport_metabolism` agent was in the `DiffusionField` schema!

```
        schema = {'agents': {}}
        for agent_id, states in self.initial_agents.items():
            location = states['boundary'].get('location', [])
            schema['agents'][agent_id] = {
                'boundary': {
                    'location': {
                        '_value': location},
                }
            }

```

This `self.initial_agents` thing to set the value is a bad practice, for just this reason. This process hangs on to a reference to `transport_metabolism` and resists all efforts to delete it! We can delete from processes and topology, but we can't delete references to things in a process's schema. 

To fix this, please initialize the location of this agent using an initial state during experiment configuration! and maybe avoid the use of `_value` at all times in schema declaration.... it is a sign we should be using `initial_state`. This should maybe just be an error now that I think of it.

vivarium/core/experiment.py
eagmon: good to know!

eagmon (commented):
Thanks @prismofeverything! Wow -- yes, ```self.initial_agents``` is bad!  I'll fix that ASAP.  Agreed on the ```_value``` schema declaration being undesirable, but the problem here was actually adding an agent in the schema!

PR #9, opened 2020-09-16
fix key in experiment update

eagmon:
The key in ```experiment``` line 1489 should be ```path```. https://github.com/vivarium-collective/vivarium-core/blob/24f2586bbee74f2d5806c978e106d951af942e6c/vivarium/core/experiment.py#L1489

### DIFF
--- a/vivarium/core/experiment.py
+++ b/vivarium/core/experiment.py
@@ -1486,7 +1486,7 @@ class Experiment(object):
             if full_step == INFINITY:
                 # no processes ran, jump to next process
                 next_event = interval
-                for process_name in front.keys():
+                for path in front.keys():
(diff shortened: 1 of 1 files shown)


U8NWXD (approved):
Good catch! When this was running before, were we getting the `path` from the previous `for` loop? I'm wondering how this got past the tests

prismofeverything:
Ha yeah, glad you found this! It never came up because this piece of code should never actually be reached in a real simulation. I have it there for completeness but the condition should never occur. Maybe I should replace it with an exception?

eagmon:
I think it's fine as is. I was just looking through the new changes to ```update``` when I found this. It didn't come up in a bug or anything like that.

PR #13, opened 2020-09-23
Register in module init, initialize_state for generators

eagmon:
This PR is to make the changes discussed in the vivarium group meeting this afternoon, plus a few more wishlist items that I have been saving up. 

*Note* -- release a new version and update all imports of plots in vivarium-cell, vivarium-chemotaxis

- [x] register register updater, dividers, and serializers in the module __init__.py
- [x] get_initialize_state for Generator
- [x] move composition plots to a `vivarium/plots/`
- [x] generalize ```timeline``` process to handle arbitrary paths into the tree.
- [ ] experiment option to print out sim time + info to terminal. This can help assure people their simulation is running.

@prismofeverything @U8NWXD what do you think of the ```initial_state``` in generator?  A ```composite``` could call each of its processes' ```initial_state``` and merge them together. Alternatively, I was also thinking about how individual processes do have their own ```generate_processes``` method, which might be of use for a more general function that iterates over each process and calls its ```initial_state```.

```Timeline's``` refactor was particularly satisfying -- we can now make a timeline that can reach into any state in the tree, by using our standard ```('tuple','path','to','state')```:

    timeline = [
        (0, {('agents', agent_id, 'global', 'dead'): False}),
        (5, {('agents', agent_id, 'global', 'dead'): True}),
        (10, {})]

I'm still thinking about a method for merging composites... Did either of you think our discussion pointed to a clear solution? My previous attempt to merge a flagella expression composite with flagella activity and receptor processes was not very successful. A more general method would be awesome.

### CODE REVIEW

File: vivarium/core/process.py [hunk 1]
```diff
@@ -116,6 +116,22 @@ def __init__(self, config):
...
+        Arguments:
+            config (dict): A dictionary of configuration options. All
+                subclass implementation must accept this parameter, but
+                some may ignore it.
+
+        Returns:
+            dict: Subclass implementations must return a dictionary
+            mapping state paths to initial values.
+        """
+        return {}
```
U8NWXD, line 133: Maybe we should have this throw an exception to force the user to override it

vivarium/__init__.py
```diff
@@ -16,3 +21,208 @@
 process_registry.register(DeriveCounts.name, DeriveCounts)
 process_registry.register(TimelineProcess.name, TimelineProcess)
 process_registry.register(NonSpatialEnvironment.name, NonSpatialEnvironment)
+
+
+# updater functions
+def update_merge(current_value, new_value, states):
```
U8NWXD, line 27: We could still define these in the registry file and import the updaters, serializers, etc. in `__init__.py`. This is mostly a stylistic preference of mine to put as little as possible in `__init__.py`

vivarium/plots/plot.py
```diff
@@ -0,0 +1,446 @@
+"""
```
U8NWXD, line 1: Should these plotting functions go into separate files? That would let us group utility functions like `order_list_of_paths` with the functions that actually use them, and that might be a more scalable approach if we add plots in the future. Also, we don't really need to have a whole `plots` folder if it's only going to contain one file. Just an idea though; I don't have strong preferences on this.

vivarium/processes/timeline.py
```diff
@@ -8,6 +8,15 @@
 from vivarium.core.process import Process, Generator
 
 
+
+def nested_set(dic, keys, value):
+    ''' make an embedded dict from a list of keys,
+    with a value set at the deepest key '''
```
U8NWXD, line 14: The first line of a docstring should stand on its own. This is what the docs will use as the short description of the function. Other lines should follow after a blank line.

vivarium/library/dict_utils.py
```diff
@@ -35,15 +35,17 @@ def deep_merge_check(dct, merge_dct):
 
 def deep_merge_combine_lists(dct, merge_dct):
     '''
-    Recursive dict merge, combines values that are lists
+    Recursive dict merge, values are lists are combined into one list without repeating values.
     This mutates dct - the contents of merge_dct are added to dct (which is also returned).
     If you want to keep dct you could call it like deep_merge(dict(dct), merge_dct)'''
```
U8NWXD, line 40: Maybe fix this docstring too while we're at it

vivarium/core/process.py
```diff
@@ -116,6 +116,22 @@ def __init__(self, config):
         if '_schema' in self.config:
             self.schema_override = self.config.pop('_schema')
 
+    def initial_state(self, config={}):
```
U8NWXD, line 119: Can we use a default of `None` instead?

vivarium/core/experiment.py
```diff
@@ -1367,8 +1367,9 @@ def emit_configuration(self):
             'experiment_id': self.experiment_id,
             'name': self.experiment_name,
             'description': self.description,
-            'processes': serialize_dictionary(self.processes),
             'topology': self.topology,
+            # TODO -- handle large parameter sets in self.processes to meet mongoDB limit
+            # 'processes': serialize_dictionary(self.processes),
```
U8NWXD, line 1372: Hmm, this might be tricky since we really do want to save all those parameters. Maybe we'll have to split them up across many documents?

eagmon: @U8NWXD -- this limit is normally sufficient. However, I ran some experiments with multiple agents that had entire E. coli genome sequences in their parameters. Maybe for now we could keep on emitting parameters, except for these large cases?

PR #14, opened 2020-09-28
Inverse topology fixes and tests

prismofeverything:
While testing support for creating topologies with partial name overrides (in anticipation of the `vivarium-bioscrape` workshop) I found a couple oversights in the `inverse_topology` implementation:

* If the update had values that weren't explicitly alongside a topology with a `_path` specification, the updates for those values were dropped.
* If the update had multiple different ports that mapped to the same key in the topology, only the last updates would be kept. 

I fixed these issues and made a more comprehensive test for `inverse_topology` to ensure this behavior is maintained. Several asserts were added. Tests again are crucial here, especially for edge cases like this and more subtle topology behaviors.


prismofeverything:
Tested in -cell, -chemotaxis, -ecoli and -bioscrape.

PR #15, opened 2020-09-29
SwapCompartment Process

eagmon:
This PR adds a general ```SwapCompartment``` process to the core processes. This takes in an existing set of processes as a parameter under the key ```'removed_processes'```, and a new compartment with replacement processes under the key ```'new_compartment'```.  If the lone state in the ```trigger``` port gets triggered to ```True``` by some other process, ```SwapCompartment ``` marks the processes in the initial compartment to be deleted, and generates the processes provided in the new compartment.

To make this process require as little information about its location in the tree, some lines in ```experiment.py``` were updated to add processes in place in the absence of a provided ```path```. A ```_generate``` update now uses an empty tuple for ```path``` if it is not provided, and ```assoc_in``` will associate the topology in-place.


eagmon:
Following @U8NWXD's suggestion, this is now ```swap_compartment``` process. This is a better description because it is exactly what it does.  Death might involve different topology updates not included here. The test at the bottom shows how the ```trigger``` port of the ```swap_compartment``` process can use a ```dead``` state if needed.

PR #16, opened 2020-09-29
Support Upload and Download Scripts

U8NWXD:
In `vivarium-scripts`, we want to be able to upload and download experiment data for archival. This adds support for these operations.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import, division, print_function
 import copy
 import numpy as np
 
+from bson.objectid import ObjectId
 from multiprocessing import Pipe
 from multiprocessing import Process as Multiprocess
 
@@ -39,6 +40,8 @@ def serialize_value(value):
             serializer_registry.access('compartment').serialize(value))
     elif isinstance(value, (np.integer, np.floating)):
         return serializer_registry.access('numpy_scalar').serialize(value)
+    elif isinstance(value, ObjectId):
+        return str(value)
     else:
         return value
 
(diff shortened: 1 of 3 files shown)


vivarium/core/process.py
```diff
@@ -39,6 +40,8 @@ def serialize_value(value):
             serializer_registry.access('compartment').serialize(value))
     elif isinstance(value, (np.integer, np.floating)):
         return serializer_registry.access('numpy_scalar').serialize(value)
+    elif isinstance(value, ObjectId):
```
eagmon, line 43: Is this how you are serializing the time value for emitted data? How would the reverse know to convert it from a string?

U8NWXD: No, this is separate from the time value. The time will automatically get serialized to a string because it's the key of a dictionary (see [`serialize_dictionary`](https://github.com/vivarium-collective/vivarium-core/pull/16/files/a07ba1a08b2a4c3a496bedf01d7d685d5c2ee6ae#diff-724eba4f1ea1557f8b74f03f93e08286R58)). For the time value, we know to convert it back to a float because it's the key of the timeseries.

This line is handling the object ID, which is automatically added by MongoDB. We actually strip the object ID when re-uploading an experiment to the database so that MongoDB can create the object IDs itself. Our code doesn't actually use the ID for anything, but if we wanted to restore it, we could since the ID is always the value of the key `_id`, and that key isn't used for anything else.

U8NWXD:
@eagmon could you take a look? I responded to your comment

eagmon (approved):

PR #22, opened 2020-10-09
Improve Code Health Infrastructure: Lint, Coverage, and Type Checking

U8NWXD:
See #42 for how to use the features introduced here


U8NWXD:
@eagmon this PR is ready to go! Could you take a look?

requirements.txt
```diff
@@ -1,14 +1,34 @@
...
-## Install requirements
-#   pip install -r requirements.txt
-#
-
--e .
\ No newline at end of file
+astroid==2.4.2
+attrs==20.3.0
+certifi==2020.6.20
+confluent-kafka==1.5.0
```
eagmon, line 4: I think `confluent-kafka` is no longer needed.  Did you get these requirements from your environment?

vivarium/processes/agent_names.py
```diff
@@ -1,12 +1,14 @@
 from __future__ import absolute_import, division, print_function
```
eagmon, line 1: We could go through and remove all these `__future__`` imports as well, since we not longer support Python 2.  Let's add that to the TODO list.

eagmon (approved):
Looks great! This will make it so much easier to keep the code health in check.

vivarium/processes/agent_names.py
U8NWXD: Done in d7f9921

requirements.txt
U8NWXD: Done in ebe99a4

requirements.txt
```diff
@@ -1,14 +1,34 @@
...
+pluggy==0.13.1
+py==1.9.0
+pylint==2.6.0
+pymongo==3.11.0
+pyparsing==2.4.7
+pytest==6.1.2
+pytest-cov==2.10.1
+python-dateutil==2.8.1
+scipy==1.5.4
+six==1.15.0
```
1fish2, line 29: Is the `six` package just an indirect dependency or is it no longer needed now that we've dropped Python 2 (RIP)?

Idea: To aid future maintenance, we could group requirements.txt into sections containing

* the direct dependencies -- anything in this group can be pruned out from here and setup.py if/when the code no longer `imports` it, like kafka
* the dev tools -- not imported; not needed in setup.py; used in dev scripts
* the frozen indirect dependencies

1fish2 (commented):
It's great to see these additional automated tests and the use of Python 3 type hints!

PR #31, opened 2020-11-02
Allow specifying timeline in agent_environment_experiment

U8NWXD:
This lets us put the timeline process at the environment level for experiments created with `agent_environment_experiment`


eagmon (approved):
looks good! So, was the timeline in each agent the source of oscillations?

U8NWXD:
Yep! That should be fixed now

PR #33, opened 2020-11-04
move update

eagmon:
This PR creates a `'_move'` update for moving nodes within the topology. `Engulf` is a new process that uses `'_move'` to take a neighbor node from its `outer` port and put it under an `inner` port. and `Burst` is a new process that uses `'_move'` to take its inner stores, merge them with its outer stores, and then delete itself.

To make this process work, I needed to make a `process_update` and `topology_update` for the `'_move'` update. To do this, a new function called `path_list_to_dict` is used to get the `process_update` embedded dict from a path list. And `Store's` `apply_update()` now gets the full `experiment_path` so that it can pull out the source node's topology. Previously, only `process_topology` was available, which wasn't enough to get any arbitrary topology for neighboring nodes.

TODO:
 - There might need to be topology-rewiring upon a node being moved, if that node is a process.
 - why is the current topology update required, with `_delete` applied later rather than earlier? Why is `self.delete_path(path)` used within the topology updates, rather than just applied with a list of `deletions`?


vivarium/core/experiment.py
```diff
@@ -638,6 +632,42 @@ def apply_update(self, update, process_topology=None, state=None):
                     target.apply_defaults()
                     target.set_value(state)
 
+            move_entries = update.pop('_move', None)
```
U8NWXD, line 635: Also, can we add `_move` to the docstring? I think there's a spot in the documentation that covers all the special keys that should also be updated

vivarium/core/experiment.py
```diff
@@ -850,24 +888,20 @@ def state_for(self, path, keys):
                 key: state.inner_value(key)
                 for key in keys}
 
-    def depth(self, path=()):
+    def depth(self, path=(), predicate=None):
```
U8NWXD, line 891: I'm not sure what `predicate` means here. Maybe `filter`?

vivarium/core/experiment.py
```diff
@@ -956,6 +990,15 @@ def establish_path(self, path, config, source=None):
             self.apply_config(config, source=source)
             return self
 
+    def add_inner(self, path, node):
```
U8NWXD, line 993: What about `add_node`? `add_inner` might be confusing since this can add to an arbitrary path, not just the current node's inner

U8NWXD (commented):
I left some suggestions. I like the current approach!

vivarium/core/experiment.py
```diff
@@ -415,6 +415,8 @@ def get_path(self, path):
 
             if child:
                 return child.get_path(path[1:])
+            elif isinstance(self.value, Process):
+                return self.value
```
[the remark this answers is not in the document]
eagmon: oops, this was something I tried earlier to get the updated process topology, but since backed out. Thanks!

vivarium/core/experiment.py
eagmon: thanks, `filter` is better.

PR #37, opened 2020-12-14
Tree updates take keys, not paths

prismofeverything:
In the process of implementing the Engulf and Burst processes, we had a realization that some confusion was arising from using paths, which are topology-relative, from inside processes, which are topology agnostic. This was leading to process configurations that modified or appended the paths, which is actually what the topology is supposed to be doing for us in the first place. 

This PR modifies all of the tree update operations to use `key` instead of `path` for their targets, which point to keys directly residing in the target ports, and updates everywhere in this repository that invokes these updates to align to the new approach. This means all path specifications will come from assigning topologies to process ports, instead of processes having to know anything about paths (which is the whole point of the topology). 

Still outstanding is updates to any other repository that references these tree related updates, which could be far reaching at this point. It is an API breaking change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ with open("README.md", 'r') as readme:
 
 setup(
     name='vivarium-core',
-    version='0.0.38',
+    version='0.1.0',
     packages=[
         'vivarium',
         'vivarium.core',
--- a/vivarium/core/experiment.py
+++ b/vivarium/core/experiment.py
@@ -662,7 +662,6 @@ class Store(object):
             if move_entries is not None:
                 # move nodes from source to target path
                 for move in move_entries:
-
                     # get the source node
                     source_key = move['source']
                     source_path = (source_key,)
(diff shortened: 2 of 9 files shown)


eagmon:
@prismofeverything -- so damn happy to see you up on the PR board again! And I'm loving this exciting update that greatly simplifies tree operations. We'll be modeling phage cycles in no time ;-).

Some questions: how close does this get us to removing `inverse_topology`, so that updates don't need to be constantly realigned with the tree structure? Also, is this ready to remove `experiment_topology` and `process_topology` function arguments to `apply_update`?

prismofeverything:
Great question. I think removing `experiment_topology` and `process_topology` are possible now, though not necessary for the scope of this feature so I am leaving it for a future PR. `inverse_topology` will be more work, but I think the ultimate outcome of this. Something comparable will have to be happening inside the tree of course, but I think it will be a more natural traversal.

PR #38, opened 2020-12-21
make a composite base class

eagmon:
This PR comes from @1fish2's suggestion to make a `Composite` base class, which subclasses `Generator`. I'll be refining this PR to more clearly separate the `Composite`-specific functions and improve documentation. Happy to incorporate any additional feedback to make this protocol as clear as possible.

Also added here -- a `register()` method for `Process`, which registers the process in the `process_registry`. This makes it easier to access processes through the registry, and seems useful for working in Jupyter/Colab notebooks.


vivarium/core/process.py
```diff
@@ -280,6 +268,41 @@ def get_parameters(self):
...
+    #
+    #     self.merge_processes = deep_merge_check(self.merge_processes, processes)
+    #     self.merge_topology = deep_merge(self.merge_topology, topology)
+    #     self.schema_override = deep_merge(self.schema_override, schema_override)
+
+
+class Composite(Generator):
+    """Composite parent class
+
+    All :term:`composite` classes must inherit from this class.
```
1fish2, line 283: Including `ToyComposite` and `InjectedGlcPhosphorylation`.

It looks like we should also add abstract classes `Compartment`, `Port`, and `Agent`.

1fish2 (commented):
Good.
I can help with this if you want.

vivarium/core/process.py
eagmon: A `Port` is an declared by the `Process` `ports_schema()` method. `Process` does also have a `self.ports()` method for getting the port names from `ports_schema()`. I'm not sure what `Compartment` and `Agent` would gain us -- they are just different applications of `Composite`. Would it be useful to have them even if they are identical at present?

1fish2: Other than the objects that define interfaces, there's no clear answer on which ones to make classes. Additional definitions entail some mental overhead, so it's wise to question the utility.

Some benefits even if the classes would be identical at present:

* Navigation: grep the code to find the Agent classes, navigate the class structure in PyCharm, and trace it at runtime. A grep for classes defined from `Generator` found classes with obvious names like `ToyCompartment`; `SplitPort` and `MergePort` which gave me the misimpression that they're Ports; `InjectedGlcPhosphorylation` which I assume is a `Composite` because it's in the composites directory; and `BB`, `PoQo`, and `MultiUnits` which I didn't look into.
* Hook: Each class is a place to add future behavior to all of its subclasses, e.g. a `Compartment` `__str__` method or an `Agent` debug feature, or to add to the interface protocol along with a default implementation. This helps more in a library with external users than in an application where we can change something and update all the relevant code.
* Documentation: Generate class hierarchy diagrams. This example is limited but I only had to right-click a class and pick Diagrams > Show Diagram Popup:
image

That's still not a clear tradeoff.

eagmon:
@1fish2 -- The one thing holding me back from merging this PR is the still unclear separation of `Composite` from `Generator`. I pulled out some functions to `Composite` -- `initial_state()` and `merge()` -- but `merge` requires `self.merge_processes` and `self.merge_topology`, which are applied in the `generate()` method. It would be nice if I could pull those instance variables out to `Composite` alone, but then I might need to re-write the `generate()`.

Also, I've been thinking -- it might be nice to add/remove processes directly to a `Process` if it inherits the `merge` method from `Generator`.  So it might make sense to move those methods back to `Generator` and leave `Composite` without any added methods.

Anyways, those are my thoughts on this at the moment. Any input is welcome.

1fish2:
Let's talk. What times work for you tomorrow, Dec. 29?

eagmon:
> Is `tuple()` different than `()`?

it appears to be the same thing.

1fish2 (commented):
I like this! Nice structure.
Sorry for getting distracted by whatever...

PR #40, opened 2021-01-02
Simplifying apply_update

prismofeverything:
Now that we are storing the topology for each process inside of the state tree and the tree altering operations have been simplified, we don't need to pass in either `process_topology` or `experiment_topology` to `apply_update`. Also, we are simplifying the use of updaters to not accept `port_mapping`s, as this could lead to inconsistent reads of state (reading state at update application time which could have diverged from update calculation time). I believe everything we were doing with `port_mapping` could be accomplished directly with updates in `next_update` or supporting derivers, and all values needed to calculate the updated state can be declared as ports and passed in with all the other state, which is guaranteed to be consistent. Happy to talk about that more and maybe to prove this assertion in all the cases `port_mapping` was used.


eagmon:
Thanks @prismofeverything! Making things simpler is harder than making things more complex, so it's great to seize on this opportunity. We'll have to work through other libraries to keep them compatible with no `port_mapping` -- I think it is just  `update_field_with_exchange` updater from vivarium-cell.

horberlan:
I would love to participate in this work, but I don't know where to start ... could you send me something here that I can use? thanks.

eagmon:
Hi @horberlan! What kind of work are you interested in? We have several example systems in: https://github.com/vivarium-collective/vivarium-cell, and our documentation is here: https://vivarium-core.readthedocs.io. If you have any particular system you are interested in setting up with Vivarium, I would be happy to help.

PR #43, opened 2021-01-07
▞▞▞ Static checkers

1fish2:
* Add runscripts for the static checkers so we don't have to remember how to run them.
  * `mypy.ini`: Add error info and name the files so you don't have to do that on the command line.
* `setup.py`: Don't require clients to install this library's static checkers.
* PEP8 errors.
  * Partial fix for too broad exception clause. We should still narrow them if possible. See https://docs.python.org/3/library/exceptions.html#exception-hierarchy
  * PEP8 is picky about the number of blank lines, variables named `l`, and such.
  * **Q.** Are there pylint settings for PEP8? I didn't see docs on what the `.pylintrc` settings do.
* **Q.** Which of these to fix vs. suppress?
  * "Variable name `plotOutDir`, `initial_A`, `A_in`, `P`, `Q`, `PQ`, `G`, ... should be lower case."
  * "Docstrings should use """Triple double-quotes""" not '''Triple single-quotes'''."
* Fix names `filter`, `format`, and `id` that shadowed built-ins.
* Put TODO comments on type errors "Type Process [or Deriver] doesn't have expected attribute 'schema'". Something's wrong but I don't know what it is.
* Start afresh with PyCharm inspection settings.
  * "Method may be static" occurs for methods that aren't overridden in this project.
  * "Shadows a name from an outer scope" and "unused local" aren't likely to lead to trouble.
* Fix mutable default args.
  * I guessed at types for those args and for a few variables that mypy needed typed. Some of these are likely to be wrong.

**Q. Which of these to fix vs. suppress before finishing this PR?**
  * "Variable name `plotOutDir`, `initial_A`, `A_in`, `P`, `Q`, `PQ`, `G`, ... should be lower case."
  * "Docstrings should use """Triple double-quotes""" not '''Triple single-quotes'''."

Q. How does import matplotlib in the top level `__init__.py` "help fix bug with import order"?

FYI, I think we can delete empty `__init__.py` files in Python 3.


runscripts/pytest.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# Run pytest tests on this project.
+# Configured by `pytest.ini` and `.coveragerc`.
+
+pytest --cov=vivarium
```
eagmon, line 5: These runscripts are a great idea. Maybe we should also add a runscript for the docs tests? I think its just `doc/test.sh ci`. Is that right, @U8NWXD?

vivarium/core/experiment.py
```diff
@@ -1349,7 +1351,8 @@ def emit_configuration(self):
             'data': data}
         try:
             self.emitter.emit(emit_config)
-        except:
+        except Exception:  # TODO(jerry): Narrow this too broad exception
+            log.exception("emitter.emit", exc_info=True, stack_info=True)
```
eagmon, line 1355: The reason I added this `try`/`except` is that some processes have enormous configs that are over mongoDB's limit -- in particular if they contain gene sequence data. So if it can't be emitted, we remove the `'process'` and `'state'` configuration data, and emit the rest.

vivarium/core/experiment.py
```diff
@@ -1423,7 +1426,7 @@ def run_derivers(self):
             if deriver:
                 # timestep shouldn't influence derivers
                 update, process_topology, state = self.process_update(
-                    path, deriver, 0)
+                    path, deriver, 0)  # TODO(jerry): Type Deriver doesn't have expected attribute 'schema'
```
eagmon, line 1429: hmm, where do you see expected `'schema'`? They should have a `port_schema()` method that gives the schema. I see the same thing below for `process` -- is this called in `process_update`?

vivarium/core/process.py
```diff
@@ -53,7 +55,8 @@ def deserialize_value(value):
     elif isinstance(value, str):
         try:
             return serializer_registry.access('units').deserialize(value)
-        except:
+        except Exception:  # TODO(jerry): Narrow this too broad exception
+            log.exception("serializer_registry", exc_info=True, stack_info=True)
```
eagmon, line 59: This is suppose to add units with the `units` serializer, only if the units have been saved to the value as a string (which happens upon serializing units).

vivarium/library/fasta.py
```diff
@@ -1,5 +1,6 @@
 from vivarium.library.path import open_relative
 
+
```
eagmon, line 3: I think this library should be moved to `vivarium-cell`. vivarium-core does not need fasta... Issue added in #44, I'll take care of this soon.

eagmon (approved):
This is excellent -- thank you @1fish2! This repo is going to be in great shape.

runscripts/pytest.sh
1fish2: Good idea. Done, and added Sphinx and sphinx-rtd-theme to `requirements.txt`.

vivarium/core/experiment.py
1fish2: Which specific exceptions can that raise? I'll change this to `PyMongoError`, which is a big improvement, e.g. it won't swallow an `AttributeError` indicating a bug.

This is probably sufficient but narrowing it further (to `WTimeoutError` and/or `DocumentTooLarge`?) would not give a pass to a bug such as a timeout trying to contact the server.

vivarium/core/experiment.py
1fish2: That message is from PyCharm's type checker. In lieu of arg type declarations on the `process_update()` method, it looks into the method implementation and sees the expression `process.schema` and thus infers the `expected attribute 'schema'` requirement.

It's either catching a bug or something weird. The only code that assigns to something.schema is in `generate_paths()` which does `subprocess.schema = subprocess.get_schema()`. OK, that's weird, and breaking encapsulation. The `Process` class has no `schema` attribute but this reaches in and adds one, and the `process_update()` method assumes it's been added.

For better code structure, it could always call `subprocess.get_schema()` and if needed that method could cache the result for future quick access, or store the computed `schema` somewhere else, or at least `Process.__init__()` should initialize a `schema` attribute.

* I'll fix it in this PR if you like.

vivarium/core/process.py
1fish2: What exception is it raising? Should it catch that exception and not log it, i.e. is it not an error?

The reason for logging is to at least not hide and swallow any arbitrary exception that this could catch. E.g. if you hit ^C it'll raise `InterruptedError` or something.

vivarium/core/composition.py
```diff
@@ -471,24 +474,35 @@ def compartment_in_experiment(
...
 
 
-
 ########################
 # simulation functions #
 ########################
 
-def simulate_process(process, settings={}):
+def simulate_process(process, settings: Optional[Dict[str, Any]] = None):
+    settings = settings or {}
```
U8NWXD, line 482: I really like this way of handling `None` defaults!

U8NWXD (approved):

vivarium/core/experiment.py
U8NWXD: Is there any reason to generate a process's schema dynamically like this? I don't think we expect to change processes after they're instantiated, so maybe a process should compute its schema only once in the constructor and then store that as an attribute. Schema computation involves a `copy.deepcopy`, so it'd be best to do it only once. Caching the result seems unnecessarily complicated though since it shouldn't change anyway

vivarium/core/process.py
U8NWXD: If a value has not been serialized, we'll get `pint.errors.UndefinedUnitError`. This isn't an error. It just means we should use the raw value instead of trying to deserialize it

1fish2: Great! I narrowed the catch to `UndefinedUnitError`.

vivarium/core/experiment.py
eagmon: `PyMongoError` sounds good to me!

1fish2:
I'm ready to merge this PR.

Note: Consider testing this `requirements.txt` file by rebuilding your virtual environment.

For future PRs:
* TODO: Configure `setup.py` to specify compatibility with Python 3.7 - 3.8.
* TODO: Make the GitHub CI test on Python 3.7 - 3.10.
* TODO: Use `_xyz` names or `__all__` declarations to choose which module-level names should be public to library clients.
* TODO: Configure pylint to suppress fewer warnings. PyCharm inspections include pylint and most of it is fine, although I suppressed 4 kinds of warnings in `.idea/inspectionProfiles/Project_Default.xml`.
* TODO: import future annotations.
* Q. `runscripts/doctests.sh` creates 16 .rst files and modifies one more. Are some or all `doc/reference/...` rst files generated files that we should `.gitignore`?
* Q. `def save_network()`'s optional argument `plotOutDir` should be `plot_out_dir` per PEP8. Would that change break any callers? For now I disabled that naming check. We can leave it as is or get back to it later.

PR #47, opened 2021-01-13
Improve the topology grapher

1fish2:
This tool is great for productivity! It draws the Process + Store topology network.

This PR tunes the rendering, mainly to show Stores using the standard flowchart symbol for data stores. The trick is to construct a matplotlib `Path` and let `draw_networkx_nodes(node_shape=s)` pass it to `ax.scatter(marker=s)`.

Labels are bigger and `..` draws as ⬆︎.

image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -51,6 +51,13 @@ exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', 'venv']
 # will help avoid broken links.
 nitpicky = True
 
+# Ignore warnings like:
+# "docstring of ... reference target not found: matplotlib.pyplot.Figure".
+# Is there a way to make those references work?
+nitpick_ignore = [
+    ('py:class', 'networkx.Graph'),
+    ('py:class', 'matplotlib.pyplot.Figure')]
+
 
 # -- Options for HTML output -------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


eagmon (approved):
I tried this out and it looks great! I'm excited to fill the colab notebook with these figures. 

I don't like that `nitpick_ignore` was required here for the matplotlib warnings! @U8NWXD -- can we ignore these warnings in the doc test? Something to talk about tomorrow in our vivarium health session.

1fish2:
Q: Is there a way to enter cross-library references in Sphinx? I'm unfamiliar with this tool.
Q: Should `/doc/reference/api/` be git ignored and deleted from the repo?

U8NWXD:
@1fish2 yes, Sphinx supports cross-library references with [`intersphinx`](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)

I think we still need `/doc/reference/api/` because while it's auto-generated, I don't think readthedocs will auto-generate those files for us. It's a command I added to `Makefile`

1fish2:
@U8NWXD How does readthedocs get the generated files? Can someone manually run Sphinx before then (and thus omit all the auto-generated files from the repo)? If not, then I guess every PR should include manually running Sphinx and committing those changes.

Whichever way, just let me know the practice to use.

PR #49, opened 2021-01-14
Code Health Fix

### LINKED ISSUE #42: Code Health
**Note**: This depends on #22.

## Code Style

* Get all code to pass lint checks
  * [x] `core/composition.py`
  * [x] `core/control.py`
  * [ ] `core/emitter.py`
  * [x] `core/experiment.py`
  * [ ]  `core/store.py`
  * [x] `core/process.py`
  * [ ] `core/registry.py`

* Complete docstrings
  * [ ] `core/composition.py`
  * [ ] `core/control.py`
  * [ ] `core/emitter.py`
  * [ ] `core/experiment.py`
  * [ ]  `core/store.py`
  * [x] `core/process.py`
  * [ ] `core/registry.py`

### How to Use Pylint

To run pylint, just execute `pylint vivarium`.

The files left to fix are listed in `.pylintrc` after `ignore=`. We use an exclusion list to ensure that once we fix a file, any code added to it also passes the linter. This prevents regressions. It also ensures that any new files added have proper style.

I use pylint's default rules. You can adjust this in `.pylintrc` after the `disable=` line. You can find the list of pylint checks [here](https://pylint.pycqa.org/en/latest/technical_reference/features.html).

### Fixing a File

1. Delete the file from the list in `.pylintrc`
2. Run pylint by executing `pylint vivarium`
3. Fix all the errors that are shown
4. Repeat steps 2-3 until pylint finds no errors

## Test Coverage

* Reach 100% code coverage
  * [x] `core/composition.py`
  * [x] `core/control.py`
  * [ ] `core/emitter.py`
  * [ ] `core/experiment.py`
  * [ ]  `core/store.py`
  * [ ] `core/process.py`
  * [ ] `core/registry.py`

### How to Calculate Coverage

I use the `pytest-cov` plugin to calculate test coverage. Whenever you run the tests, add the `--cov=vivarium` argument to calculate coverage. At the end of the test output, you'll get a table like this:

```
Name                                                  Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------------------
vivarium/__init__.py                                     37      0   100%
vivarium/composites/__init__.py                           0      0   100%
vivarium/composites/injected_glc_phosphorylation.py      13      0   100%
vivarium/core/__init__.py                                 0      0   100%
vivarium/core/composition.py                            403    180    55%   62, 64-68, 141, 143, 156-157, 179-193, 200-221, 247-333, 347, 364-370, 375-386, 417, 445-456, 527-540, 543-544, 548-552, 559-569, 578-580, 614-643, 686-706, 746-755, 998-1000
vivarium/core/control.py                                 85     13    85%   40, 42, 59-60, 63-64, 89-90, 104-106, 112, 119, 211
vivarium/core/emitter.py                                156     80    49%   32-33, 55, 59, 65, 70-73, 86-94, 103, 116, 119, 128, 176-189, 192-196, 199, 203-212, 216-222, 226-227, 232-280, 284-298, 304-312, 320-324
vivarium/core/experiment.py                            1018    104    90%   93, 102, 171, 173, 182-184, 191, 245, 250-251, 262, 267, 271-272, 285-286, 288, 293, 311, 327-328, 381-384, 416, 437, 442, 447, 452, 460-467, 481, 485, 487, 490, 500-502, 520-523, 665-672, 827-831, 855-872, 909, 920, 922, 948-949, 990-997, 1014, 1069, 1076, 1096, 1348-1352, 1461-1462, 1494-1495, 1517, 1546, 1548, 1570-1576, 1677, 2254
vivarium/core/process.py                                277     40    86%   30, 41, 43, 51, 78-81, 100-104, 211, 227, 274, 277-279, 319, 337, 353-361, 370, 386, 392-394, 401, 457, 459, 507, 516, 564
vivarium/core/registry.py                               118     57    52%   77-78, 115-122, 146-153, 164, 182-197, 202-204, 212, 227, 238, 246, 249, 253, 256, 260-264, 271-275, 284-286, 289, 294-296, 300
vivarium/experiments/__init__.py                          0      0   100%
vivarium/experiments/glucose_phosphorylation.py          48      5    90%   91-96, 105
vivarium/library/__init__.py                              0      0   100%
vivarium/library/datum.py                                28     19    32%   2-3, 6-7, 28-41, 44, 47, 50
vivarium/library/dict_utils.py                          163     75    54%   17-20, 33-36, 38, 53, 68, 70, 74, 94, 113-117, 127-132, 136-144, 150-152, 156-175, 183-202, 206, 230-236, 259-261
vivarium/library/fasta.py                                10      8    20%   4-12
vivarium/library/filepath.py                             35     22    37%   22-25, 31-34, 51-58, 73-77, 81-82, 87-89, 93-94
vivarium/library/make_network.py                         87     79     9%   15-27, 33-40, 55-92, 96-125, 136-171
vivarium/library/path.py                                  4      2    50%   4-5
vivarium/library/pretty.py                               24      5    79%   16, 18, 20-21, 26
vivarium/library/schema.py                               39     28    28%   4, 7, 12, 18-25, 28-40, 45-52, 55, 62, 69, 77
vivarium/library/timeseries.py                           32     27    16%   27-50, 61-69
vivarium/library/topology.py                            148     17    89%   36-37, 58-61, 93-98, 107-116
vivarium/library/units.py                                38      6    84%   62-66, 90
vivarium/plots/__init__.py                                0      0   100%
vivarium/plots/agents_multigen.py                       135    127     6%   13-24, 64-226
vivarium/plots/simulation_output.py                     103     16    84%   70, 76, 78-79, 82, 84, 99-103, 122, 129, 132, 138, 172
vivarium/plots/topology.py                              105     17    84%   40, 137-140, 151, 166-180, 187, 202, 255, 268-277
vivarium/processes/__init__.py                            0      0   100%
vivarium/processes/agent_names.py                        14      6    57%   12-14, 17, 29-30
vivarium/processes/burst.py                              58      6    90%   208-212, 217
vivarium/processes/derive_concentrations.py              28     19    32%   20-31, 34, 53-69
vivarium/processes/derive_counts.py                      24      8    67%   32, 35, 55-62
vivarium/processes/divide_condition.py                   15      7    53%   10-11, 14, 17, 25-28
vivarium/processes/engulf.py                             58      6    90%   183-187, 191
vivarium/processes/exchange_a.py                         15      0   100%
vivarium/processes/glucose_phosphorylation.py            39      5    87%   129-139
vivarium/processes/injector.py                           37      9    76%   57, 103-109, 113
vivarium/processes/meta_division.py                      76      9    88%   29, 52, 65, 207-211, 215
vivarium/processes/nonspatial_environment.py             27     15    44%   22-24, 27-72, 75-87
vivarium/processes/remove.py                             47      6    87%   116-120, 124
vivarium/processes/swap_processes.py                     63      6    90%   175-179, 183
vivarium/processes/template_process.py                   38      7    82%   138-146, 151
vivarium/processes/timeline.py                           57      5    91%   43-44, 51, 56-57
vivarium/processes/tree_mass.py                          50      7    86%   69, 168-172, 176
-----------------------------------------------------------------------------------
TOTAL                                                  3752   1048    72%

Required test coverage of 72.0% reached. Total coverage: 72.07%
```

This lists the fraction of lines in each file that are covered by the tests. It also tells you which lines in the file weren't covered so you can add tests for them.

Notice the last line says we require a code coverage of 72%. This is just the current coverage value. If coverage dips below 72%, the tests will fail to make sure we don't regress.

### How to Improve Coverage

1. Run the tests with `pytest --cov=vivarium`.
2. Pick a file that is not completely covered (coverage < 100%) using the coverage output.
3. Match the lines listed as not being covered to the source file. These are the lines you need to write tests for.
4. Add tests that will check that the uncovered lines are working correctly.
5. Run the tests again. You should get a coverage value of 100% for that file.
6. Update the minimum coverage in `.coveragerc` to your current coverage value for the entire project.

## Static Typing

* Annotate all code with types and resolve all issues raised by mypy
  * [x] `core/composition.py`
  * [x] `core/control.py`
  * [ ] `core/emitter.py`
  * [ ] `core/experiment.py`
  * [x] `core/process.py`
  * [ ] `core/registry.py`

### How to Use Mypy

We use type annotations as specified by [PEP 484](https://www.python.org/dev/peps/pep-0484) and prefer annotations to special comments like `# type: str`. See the [documentation for `typing`](https://docs.python.org/3.8/library/typing.html) for more information.

Mypy is configured using the `mypy.ini` file, which looks like this:

```ini
[mypy]

disallow_untyped_defs = True
disallow_incomplete_defs = True

# Ignore missing types from third-party libraries.
[mypy-numpy.*]
ignore_missing_imports = True
...

# Ignore missing types from Vivarium Core files we haven't typed yet.
[mypy-vivarium.core.registry.*]
disallow_untyped_defs = False
disallow_incomplete_defs = False

[mypy-vivarium.core.control.*]
disallow_untyped_defs = False
disallow_incomplete_defs = False
...
```

Notice how we first set `disallow_untyped_defs` and `disallow_incomplete_defs` to True. This means that by default, mypy will raise an error whenever a function is not completely typed. Then, for every file that hasn't been typed yet, we include a section like `[mypy-vivarium.core.registry.*]` for `vivarium/core/registry.py` that sets those options to False. As we add type hints, we can remove files from this list until everything has type annotations.

Also note that we tell mypy to ignore types in our dependencies like `numpy`. Many third-party libraries are poorly-typed, and we want to avoid spending time trying to get mypy to work with them. We will focus on making sure the types in our code are correct instead.

### How to Add Types

1. Pick a file listed in `mypy.ini` as not being covered yet.
2. Delete the file's section from `mypy.ini`.
3. Run mypy by executing `mypy vivarim`
4. Fix any errors raised by mypy
5. Repeat 3-4 until no errors are raised

U8NWXD:
Fixes #42 

We claim files in a [spreadsheet](https://docs.google.com/spreadsheets/d/1AWParvAVmcHNODLmzYiSi9svHsfO6LH0V-9XVqAdqMc/edit?usp=sharing).


U8NWXD:
@eagmon We have a few `TODO` comments throughout the codebase. What do you want to do with them? We could turn them into github issues, for example. I'm not sure what they all mean or if they're even still relevant, so maybe we should go through them all at some point and decide which we actually need to act on

eagmon:
@U8NWXD -- ignoring TODOs in pylint is good for now. I'll look through and remove/resolve the easy ones in this PR -- adding it to the spreadsheet.

eagmon:
We now have a `composites/toys.py` with many toy Processes and Composites that can be reused.

eagmon:
@prismofeverything -- Since you are signed up for experiment test coverage you should know I moved `experiment.py` tests to a new file `test_experiment.py`, and some of it to `composites/toys.py`. Perhaps `test_experiment.py` should be moved out of `core/`. This reduced `experiment.py` by nearly 1000 lines! We've been doing great with tests there.

U8NWXD:
@1fish2 I saw you added `no-self-use` to the list of disabled checks. I also encountered issues with that in abstract methods that could be overridden to use `self`, but I disabled `no-self-use` on just those lines since in general, I think it's good to make methods static where possible. What do you think?

U8NWXD:
Also why do we have `pull_data`? It doesn't look like we ever override it:

https://github.com/vivarium-collective/vivarium-core/blob/758a97aa6bda333330bc2bc6ff6ecd0077631814/vivarium/core/process.py#L411-L412

1fish2:
`no-self-use`: Good question.

There's a design decision to make on each method: make it a class method or an instance method? That decision goes well beyond whether it happens to refer to `self`, e.g. whether subclasses are allowed to override it (that's the point of `is_deriver(self)`), whether it's _about_ the instance or the class, whether it's used to create an instance without deriving it from another one, etc.

pylint is in no position to detect that design decision.

So my recommendation is to disable this pylint check and then we don't need the comments to locally disable it.

(This is putting aside the question of whether to remove/deprecate the "deriver" feature. Also, one can override a Python class method but that'll have subtle semantic differences.)

In a library with a wide outside world of possible clients, we can't change design decisions like these without breaking other people's programs. Before publishing the Vivarium paper we have to:

* finalize some decisions like which methods are class methods
* hide names that are intended to be private via the `__all__` technique or the `_xyz` name technique
* hide modules that are intended to be private

1fish2:
`pull_data()` what is this method for? It is called from `vivarium/core/experiment.py`.

With the limited number of library clients outside our team so far, can we find out if we can safely remove any particular method or feature?

U8NWXD:
`TypedDict` was added in Python 3.8. I used it in `process.py`, which is why the tests for Python 3.6 and 3.7 are failing. Do we need to keep supporting 3.6 and 3.7? My usage of `TypedDict` was pretty minor, so I'm fine either way

U8NWXD:
For anyone working on adding type annotations, I created some types you can use:

```python
Path = Union[Tuple[str, ...], Tuple[()]]
Topology = NewType('Topology', dict)
Schema = NewType('Schema', Dict[str, Any])
State = NewType('State', Dict[str, Any])
Update = NewType('Update', Dict[str, Any])


class CompositeDict(TypedDict):
    processes: Dict[str, 'Process']
    topology: Topology
```

The code is [here](https://github.com/vivarium-collective/vivarium-core/pull/49/files#diff-849679ee8b6c3407c3fd17e3211ca7b1c2fb4a384ff3628d1f55bfef48fccbbeR28-R32).

eagmon:
@1fish2 -- `pull_data()` is used to emit information about a Process if it is flagged with `self.emit=True`, though I don't think it is ever used. We put it in there for the `wcecoli_process` in the wcEcoli repo, but no longer use it there either. Now we are just getting all the relevant data within that process' `next_update`, which is emitted as `Store` values. I think we should keep it just incase anyone wants to call emit on a Process.

eagmon:
@U8NWXD -- can you put the type annotations in a library perhaps? I believe there are files that can't import from Process due to circular imports.

eagmon:
@1fish2 -- "can we find out if we can safely remove any particular method or feature?" -- it would be very useful if we could. There are several dependent libraries now, and I believe it is only us or close collaborators.

U8NWXD:
> @U8NWXD -- can you put the type annotations in a library perhaps? I believe there are files that can't import from Process due to circular imports.

Done @eagmon! I also removed `TypedDict` for now

U8NWXD:
btw I am working on a fix for the documentation failures. I need to add some code to help Sphinx resolve type references

vivarium/core/types.py
```diff
@@ -0,0 +1,9 @@
+from typing import Union, Tuple, NewType, Dict, Any
+
+
+Path = Union[Tuple[str, ...], Tuple[()]]
+Topology = NewType('Topology', dict)
+Schema = NewType('Schema', Dict[str, Any])
+State = NewType('State', Dict[str, Any])
+Update = NewType('Update', Dict[str, Any])
+CompositeDict = NewType('CompositeDict', Dict[str, Any])
```
1fish2, line 9: Yay for defining these! Named concepts will greatly assist people learning to use Vivarium. The type declarations also serve as very concise reference docs on how to build these values.

* Add `Config = Dict[str, Dict[str, Any]]`?
* Can we make these more specific than `dict` and `Dict[str, Any]`?
* Do we really want these to be `NewType` types or just type aliases? mypy treats a `NewType` like a distinct subclass, which means it auto-coerces to its base class but not vice versa. (The term "newtype" comes from Haskell where it creates a distinct new type that won't coerce to/from any other type. In various other languages, a struct type is distinct from another struct type even if it happens to contain the same element types. That's a "newtype.") So if `Schema` is a `NewType`, then all the code to construct Schemas has to call the "constructor" (a no-op function), e.g.: `Schema({'k1': 223, '_default': 0})` or `Schema(dict(k1=223, _default=0))` or something has to explicitly "convert" them.
* If it's fine to call constructors when building each of these values, then why not make them real `dict` subclasses (nominal types)? Then we could add methods like `__str__()`, debugging aids like optional runtime checks, places where you can put a breakpoint, etc.
  * Actually, Factory could auto-convert a `dict` to a `Schema`, `Config`, etc., allowing the caller to use the dict literal syntax while the rest of the code passes around actual `Schema` objects to aid debugging and type checking. In that case Factory would accept a `Union[Schema, Dict[str, Any]]` while most of the library code would accept `Schema`.
* There's a backward-compatible syntax for `TypedDict` so we can still use it to get mypy to type-check common dict items, e.g. that `_updater` holds a `Literal['accumulate', 'set', 'merge']`. A simpler, optional approach is to add utility functions, like say (rough idea):
  ```
  def subconfig(default: Any, updater: Literal['accumulate', 'set', 'merge'], **kwargs):
      return dict(kwargs, _default=default, _updater=updater)
  ```

U8NWXD:
I've removed the autodoc stubs from `doc/reference/api`. They are now autogenerated by Sphinx whenever we build the docs

U8NWXD:
@eagmon I'm seeing the following lint errors from `composition.py` and `composites/toys.py`:

```
************* Module vivarium.core.composition
vivarium/core/composition.py:432:0: C0301: Line too long (84/80) (line-too-long)
vivarium/core/composition.py:385:4: W0612: Unused variable 'experiment' (unused-variable)
vivarium/core/composition.py:392:4: W0612: Unused variable 'experiment' (unused-variable)
vivarium/core/composition.py:10:0: W0611: Unused Composite imported from vivarium.core.process (unused-import)
************* Module vivarium.composites.toys
vivarium/composites/toys.py:348:8: R1725: Consider using Python 3 style super() without arguments (super-with-arguments)

-------------------------------------------------------------------
Your code has been rated at 9.95/10 (previous run: 10.00/10, -0.05)
```

Are you working on those files? I can go ahead and fix those errors if you aren't

vivarium/core/types.py
```diff
@@ -0,0 +1,9 @@
+from typing import Union, Tuple, NewType, Dict, Any
+
+
+Path = Union[Tuple[str, ...], Tuple[()]]
```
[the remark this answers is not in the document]
U8NWXD: I've updated the type alias to just `Tuple[str, ...]` and renamed the type to `HierarchyPath` in 7d538b4

eagmon:
@U8NWXD -- go ahead and fix the lint errors, thanks!

vivarium/core/types.py
U8NWXD: Re making types more specific: I think we're blocked by the fact that mypy [doesn't support recursive types](https://github.com/python/mypy/issues/731). Given that constraint, I don't see how to make them any more specific. I'm open to doing that though if you see a way!

Re NewType vs Class vs Alias: I agree with making these types into full classes eventually. Since mypy treats them as distinct types, we have to treat the type like a class anyway like this:

```python
def generate_topology(self):
    return Topology({})
```

Therefore it should be really easy to replace the `Topology` type with a full class later on if we want to. This is why I used NewType instead of just a type alias

1fish2:
>
>  Re making types more specific: I think we're blocked by the fact that
> mypy doesn't support recursive types
> <https://github.com/python/mypy/issues/731>. Given that constraint, I
> don't see how to make them any more specific. I'm open to doing that though
> if you see a way!
>
> Re NewType vs Class vs Alias: I agree with making these types into full
> classes eventually. ... Therefore it should be really easy to replace the
> Topology type with a full class later on if we want to. This is why I
> used NewType instead of just a type alias.
>

Aha. I was confusing a topology spec (et al) that clients pass in with a
topology structure that the library deals with internally.

The latter can be an ordinary class, maybe subclassing dict or not, and we
use that class name. No mypy recursion problem. These types help us
understand and cross-check the library code. They can be NewTypes for now
or go ahead and define the classes soon, perhaps after the first pass of
this PR work.

The former types are important for users to learn, understand, and
type-check their code. These are the ones to have more details such as what
I wrote for Config.

Name them like Topology and TopologySpec? TopologyImpl and Topology? Or...?

1fish2:
... and if TopologySpec wants to be recursive, just use Any or object in
that spot with a comment.

>

eagmon:
We now have a new core file, `tree.py`, for the `Store` object that was previously in `experiment.py`. This brings `experiment.py` down to about 500 lines -- so those tests can reasonably go back to `experiment.py` now... `tree.py` is about 1000 lines. I think these are more reasonably-sized, and within the pylon default limit. @prismofeverything -- I added tree.py to our spreadsheet, and added your name under what you had on experiment.py.

U8NWXD:
> Aha. I was confusing a topology spec (et al) that clients pass in with a
topology structure that the library deals with internally.

> The latter can be an ordinary class, maybe subclassing dict or not, and we
use that class name. No mypy recursion problem. These types help us
understand and cross-check the library code. They can be NewTypes for now
or go ahead and define the classes soon, perhaps after the first pass of
this PR work.

I'm not seeing what this distinction buys us. Take `Topology` for example, it needs to support key-value pairs of strings as well as arbitrary nesting. This means all of the following are valid topologies:

```python
topo = {
    "a": "b",
}
topo = {
    "a": {
        "b": "c",
    },
}
topo = {}
```

We'd like to specify that the topology, both the "topology spec" and the "topology structure," should have strings as keys, and either strings or topologies as values. This requires recursive types.

> The former types are important for users to learn, understand, and
type-check their code. These are the ones to have more details such as what
I wrote for Config.

@1fish2 I'm assuming you are referring to this type for Config: `Config = Dict[str, Dict[str, Any]]`. I don't think this config will work because it doesn't allow top-level values to be strings, which we often use. For example, this config wouldn't be allowed:

```python
config = {
    'key': 'val',
}
```

U8NWXD:
After discussion with @eagmon, I'm going to make my types type aliases instead of NewTypes to reduce the difficulty of this migration

1fish2:
> I'm not seeing what this distinction buys us. Take `Topology` for example, it needs to support key-value pairs of strings as well as arbitrary nesting. This means all of the following are valid topologies:
> 
> ```python
> topo = {
>     "a": "b",
> }
> topo = {
>     "a": {
>         "b": "c",
>     },
> }
> topo = {}
> ```

To me, these look like specifications of networks to build, in contrast with run-time objects that would have methods to walk and edit the network of Processes and Stores. That's my conceptual distinction.

Maybe Vivarium doesn't need those run-time objects since it deals directly with the constructed Processes and Stores, or maybe those run-time objects are Compartments? Apologies for not having time to fully read the docs! This is why I'd like to discuss it in a meeting.

> We'd like to specify that the topology, both the "topology spec" and the "topology structure," should have strings as keys, and either strings or topologies as values. This requires recursive types.

Type systems are limited, so one approach is to use `Any` (with a comment) in a nested slot that should be a recursive type reference.

Another approach is to use a named class which does support recursive types:

```python
from __future__ import annotations

from typing import List, Optional

class Tree:
    def __init__(self, left: Optional[Tree], right: Optional[Tree]) -> None:
        super().__init__()
        self.left: Optional[Tree] = left
        self.right: Optional[Tree] = right
        self.subs: List[Tree] = []
        if left:
            self.subs.append(left)
        if right:
            self.subs.append(right)
```

(Without `__future__ annotations`, the arg types need to be `Optional['Tree']` to handle forward references to the not-yet-defined class but the instance variable types needn't. You can see why this future is an improvement.)

> > The former types are important for users to learn, understand, and
> > type-check their code. These are the ones to have more details such as what
> > I wrote for Config.
> 
> @1fish2 I'm assuming you are referring to this type for Config: `Config = Dict[str, Dict[str, Any]]`. I don't think this config will work because it doesn't allow top-level values to be strings, which we often use. For example, this config wouldn't be allowed:
> 
> ```python
> config = {
>     'key': 'val',
> }
> ```

Isn't a config a dict of process name -> process parameter name -> process parameter value? Or is that only a Composite config? Some of the types I put in were straw-mans for feedback.

Nominal types or type aliases could help document these things more clearly like maybe:

```python
ProcessName = str
ProcessParameterName = str
ProcessParameterValue = Any
ProcessConfig = Dict[ProcessParameterName, ProcessParameterValue]
CompositeConfig = Dict[ProcessName, ProcessConfig]
```

U8NWXD:
Two questions:
1. Processes support self-registering when given a `_register` key in their parameters. Do we still use this, or can it be removed?
2. Should `generate_processes` and `generate_topology` be private methods? I think we'd always call `generate()` instead of the 2 other functions, right?

U8NWXD:
It looks like `from __future__ import annotations` doesn't work in python 3.6 or 3.7. Should we drop support for those versions?

eagmon:
@U8NWXD -- I think it's okay to drop support for 3.6 and 3.7.

eagmon:
@U8NWXD -- I dropped support for 3.6 and 3.7 in the GitHub workflows with c9f8257

eagmon:
@U8NWXD -- have we decided what to do about test coverage for items under `if __name__ == '__main__':`? It would be nice to keep them running, but they are not covered by tests. You mentioned the possibility of somehow marking them individually to be ignored.

1fish2:
Per [pylint #1970](https://github.com/PyCQA/pylint/issues/1970), the tool to do PEP8 checks is [pycodestyle](https://github.com/PyCQA/pycodestyle). The fact that pylint checks some PEP8 details is historical.

eagmon:
I added "pragma: no cover" to the `exclude_lines` in `pytest.ini`. This can now be used to selectively ignore lines from the pytest coverage test by adding it in a comment, as is needed for anything under `if __name__ == '__main__':`:
```
test_composite_in_experiment()  # pragma: no cover
```

PR #69, opened 2021-04-06
custom widths in topology plots

WilliamIX:
This PR adds a new keyword to the plot_topology function: custom_widths, which allow each process or store to have custom widths when plotted via networkx.


eagmon (approved):
Looks great! Thanks @WilliamIX

PR #70, opened 2021-04-10
Provide information about update failure

prismofeverything:
Currently if the updater for an update ever ends up being `None`, it fails with a cryptic message about not being able to apply None as a function. This PR detects the failure and provides useful information about what went wrong: 

* the path to this node
* its current value
* what the update is that was trying to be applied, but whose updater is `None` for whatever reason. 

Beyond this case I think this is a general usability principle, that we should be on the lookout for cryptic failures that take awhile to figure out (especially things deep inside store) and figure out a way to provide more information to the user about what might be going wrong (!)

### DIFF
--- a/vivarium/core/store.py
+++ b/vivarium/core/store.py
@@ -8,6 +8,7 @@ The file system for storing and updating state variables during an experiment.
 
 import copy
 import logging as log
+from pprint import pformat
 
 import numpy as np
 from pint import Quantity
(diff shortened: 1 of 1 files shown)


eagmon:
Hey @prismofeverything -- this is great, thank you! We should aim to have many more exceptions with information like this. Perhaps we can open an issue, and have people drop comments about what information would have been useful for their modeling experiences.

PR #91, opened 2021-09-14
Alternator process

prismofeverything:
This PR creates a new `Alternator` process that cycles around a given set of `choices` where at most one is True, and remains in each state for a given set of `periods`. 

```
alternator = Alternator({
    'choices': ['ON', 'OFF'],
    'periods': [5.0, 3.0]})
```

will generate a state of alternating

```
{'time': 0.0, 'ON': True, 'OFF': False}
{'time': 5.0, 'ON': False, 'OFF': True}
{'time': 8.0, 'ON': True, 'OFF', False}
```

If there are more choices than periods, it will cycle through all given periods before repeating them (essentially modulus the choice index). 

To facilitate this I added a new key to the fundamental `Process` parameters called `_condition`, which specifies a new port key pointing to a boolean state variable that determines whether or not it should run. This way any `Process` can become conditional simply by passing this key as a parameter (much like `_parallel`), and its condition can be modulated by any process writing to its condition variable, which can be wired anywhere else in the `Store` according to the provided topology.


vivarium/core/engine.py
```diff
@@ -500,8 +511,10 @@ def calculate_update(
             ``path``) and the store at ``path``.
         """
         store, states = self.process_state(path, process)
-        return self._process_update(
-            path, process, store, states, interval)
+        if process.update_condition(interval, states):
+            return self._process_update(
+                path, process, store, states, interval)
```
eagmon, line 516: Is this the right place for `process.update_condition` -- what if a different `Invoke` is used?

prismofeverything: This is the update condition for derivers, the other one is in the main `update` loop. I believe it is correct? The different invoke should still work from this point, and not be invoked if the condition is not met.

eagmon: Should this `update_condition` come back then?: https://github.com/vivarium-collective/vivarium-core/blob/f68996dc53c41376a2173a8db0b040082a30a120/vivarium/core/engine.py#L124

prismofeverything: Okay just verified `Invoke` is used inside the `update_condition` block, which is what we want. Also verified that derivers can be parallelized as well.

PR #93, opened 2021-10-01
notebook tutorials to vivarium-core docs

eagmon:
This feature will add the vivarium_interface_basics notebook to the docs tutorial page, and make it easier for future notebooks to be added as well.


.gitignore
```diff
@@ -22,6 +22,8 @@ user/*
 ###################
 *.pyc
 __pycache__
+/doc/reference/api/
```
U8NWXD, line 25: Why are we adding this? I think when we changed the vivarium-core docs to be more hand-crafted we stopped auto-generating the api stubs

doc/conf.py
```diff
@@ -163,4 +178,4 @@ def resolve_intersphinx_aliases(app, env, node, contnode):
 def setup(app):
     app.connect('doctree-read', resolve_internal_aliases)
     app.connect('missing-reference', resolve_intersphinx_aliases)
-    app.connect('autodoc-skip-member', autodoc_skip_member_handler)
+    app.connect('autodoc-skip-member', autodoc_skip_member_handler)
```
U8NWXD, line 181: Can you add the newline at the end back? I think it's usually a best-practice to always end files with a newline

U8NWXD (changes requested):

PR #96, opened 2021-10-14
Implement Topology Caching

thalassemia:
Profiling of `vivarium-ecoli` revealed that major time savings can be had by reducing the amount of time spent in the `schema_topology` function. One way to accomplish this is by caching the topology so that it does not have to be filled in again for timesteps where it does not change.

Here is a rudimentary caching implementation that only works if the topology never changes. I played around with `functools.lru_cache`, but it doesn't work OOTB because dictionaries are an unhashable type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,10 @@ out/
 user/*
 !user/user.txt
 *.cprof
+.env
+Pipfile
+Pipfile.lock
+.vscode/*
 
 # Compiled source #
 ###################
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -19,6 +19,7 @@ more-itertools==8.5.0
 nbsphinx==0.8.7
 networkx==2.5
 numpy==1.19.1
+orjson==3.6.4
 packaging==20.4
 Pillow==7.2.0
 Pint==0.15
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ numpy==1.19.4
 Pint==0.16.1
 pymongo==3.11.0
 scipy==1.5.4
+orjson==3.6.4
 
 # dev tools
 mypy==0.790
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ if __name__ == '__main__':
             'Pint',
             'pymongo',
             'scipy',
+            'orjson'
         ],
         classifiers=[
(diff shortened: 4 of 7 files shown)


thalassemia:
@U8NWXD The [StackOverflow thread](https://stackoverflow.com/a/22003440) we were referencing actually had a better, faster solution to the nested dictionary hashing problem: serializing to json. 

I tried implementing `lru_cache` using the standard library json module in vivarium-ecoli, but found it to be painfully slow. I'll link the flamegraphs below, but the gist of it is that converting to json ends up being slower than just calling schema_topology sans caching.

Instead, what I've implemented in 2671ef5 makes use of the much faster `orjson` package to perform serialization and deserialization. The upshot is that it really is much faster than standard `json` and can even natively serialize numpy datatypes. The downside is it brings in an additional dependency (which I've added to our `requirements.txt` files to make the CI checks pass). Neither `json` nor `orjson` can handle `Quantity` and function objects by default, so I also created a naïve serializer for those objects of those types which simply returns their string representations. 

As an aside, when implementing this solution in vivarium-ecoli, I was running into a weird bug where `orjson.dumps()` complained about non-string dictionary keys even though I manually confirmed that every key in the relevant dictionary was indeed a string. To get around this, I'm running `orjson.dumps()` with the option to serialize non-string keys, which is slower and not necessarily stable (if you, for instance, have both 1 and "1" as keys). Ideally, I'll figure this out and we won't need to use the non-string option going forwards.

Here are the relevant flamegraphs of 100 second vivarium-ecoli simulations using `orjson`-based LRU caching, `json`-based LRU caching, and no caching: [Google Drive folder](https://drive.google.com/drive/folders/1glun_7jDprpIw6rnhad6a6wxA9hi5F4C?usp=sharing)

eagmon:
Thank you @thalassemia and @prismofeverything for the discussion. I finally caught up -- this was not something I want to skimp on. It sounds like @prismofeverything's solution might even be better performance-wise, but would require that we track everywhere that topology is changed. Or is it possible to trigger a recache every time topology is altered without tracking those methods?

prismofeverything:
I am pretty sure the topology is only changed in two places, so I don't think it's a huge change. That said! This is already a huge improvement performance-wise. Do we need a quick call to come to a consensus here perhaps?

PR #98, opened 2021-10-23
Add Steps for Parallelized "Derivers"

U8NWXD:
To run large simulations with complex derivers, we want to be able to run derivers in parallel. This means that we need a dependency graph instead of a sequential list to specify the order in which to run derivers. This PR introduces steps to replace derivers and support this dependency graph structure. More specifically, this PR does the following:

* [x] Replace derivers with steps
* [x] Manage steps as a dependency graph
* [x] Execute steps in layers (topological generations)
* [x] Allow steps in the same layer to run in parallel
* [x] Let the user specify tasks and task dependencies as part of a composite. This should also handle division.

The following tasks are left for a later PR:

* [ ] Skip steps based on whether states have been updated. When skipping a task, also skip any tasks that depend on it.
* [ ] Allow OR dependencies (i.e. dependency relationship of the form "run step1 if step2 or step3 have run."

Questions:

* This PR adds new functionality, so should we bump the minor version instead of the patch?
  * No. We bump the minor version for breaking API changes only.
* This PR does break some public APIs, for example `Store.generate()`. However, these public APIs aren't often used since we usually run experiments through Engine, so does this count as a breaking change?


vivarium/core/composer.py
```diff
@@ -304,36 +369,47 @@ def __init__(
         super().__init__(config)
         self.composers: List = list(composers)
 
+    def _generate(
+            self, config: Optional[dict], method: str
```
eagmon, line 373: nice use of `method`!

vivarium/core/engine.py
```diff
@@ -270,22 +386,32 @@ def __init__(
...
+            self.steps = self.state.get_steps()
+            self.flow = self.state.get_flow()
             self.topology = self.state.get_topology()
         else:
             raise Exception(
                 'load either store or (processes and topology) into Engine')
 
+        steps = steps or {}
+        flow = flow or {}
+
```
eagmon, line 414: Why not use `self.steps`, `self.flow`?

vivarium/core/engine.py
```diff
@@ -337,23 +465,66 @@ def __init__(
         log.info(pf(self.topology))
 
 
+    def _add_step_path(
+            self,
+            step: Step,
+            path: HierarchyPath,
+            # None if deriver, empty list if no dependencies.
+            relative_dependencies: Optional[Sequence[HierarchyPath]],
+            depend_on_derivers: bool = False,
```
eagmon, line 474: What is this `depend_on_derivers` used for? It looks like deriver-like steps are added in `_add_process_path` by not having `relative_dependencies`.

vivarium/core/engine.py
```diff
@@ -541,7 +716,14 @@ def apply_update(
...
             for path, process in process_updates:
                 assoc_path(self.processes, path, process)
-                self._add_process_path(process, path)
+                self._add_process_path(process, path, {})
+
+        if step_updates:
+            for path, step in step_updates:
+                dependencies = flow_update_dict.get(path)
+                assoc_path(self.steps, path, step)
+                self._add_step_path(step, path, dependencies)
```
eagmon, line 725: Is this tested anywhere? It would be good to test that division is applied correctly with `step_updates`.

vivarium/core/engine.py
```diff
@@ -563,32 +745,43 @@ def delete_path(
...
+            deferred_updates: List[Tuple[Defer, Store]] = []
+            for path in layer:
+                step = self.step_paths[path]
+                if not step:
+                    # Step was deleted by a previous step.
+                    continue
+                # Timestep shouldn't influence steps.
                 # TODO(jerry): Do something cleaner than having
                 #  generate_paths() add a schema attribute to the Deriver.
                 #  PyCharm's type check reports:
```
eagmon, line 772: does this still apply?

vivarium/core/process.py
```diff
@@ -323,21 +346,24 @@ def update_condition(
...
+class Step(Process, metaclass=abc.ABCMeta):
+    """Base class for steps."""
 
-         Given a config, this method should return the Deriver's initial
-         state.
-         """
-        return {}
+    def __init__(self, parameters: Optional[dict] = None) -> None:
+        parameters = parameters or {}
+        self.dependencies = parameters.pop('dependencies', [])
```
eagmon, line 354: is `Step.dependencies` still needed?

vivarium/core/store.py
```diff
@@ -1023,15 +1245,29 @@ def divide(self, divide):
...
                 topology = topology or {}
 
+            # get the daughter flow
+            if 'flow' in daughter:
+                flow = daughter['flow']
+            else:
+                # if no flow provided, copy the mother's flow
+                mother_flow = self.get_path(mother_path).get_flow()
+                flow = copy.deepcopy(mother_flow)
+                flow = flow or {}
```
eagmon, line 1255: looks good!

eagmon (approved):
This is a really exciting new feature! Just a few comments/questions below. 

I think we can bump the version to 0.4.0 given the minor API changes. I don't think any of the dependent libraries use the breaking API, but might as well keep up that policy.

vivarium/core/engine.py
U8NWXD: This describes whether the first step should be added with a dependency on any existing derivers. Now that I'm writing this, I realize that we need to to the topological sorting first so that we know which step(s) come first. Supporting derivers this way might be difficult...

vivarium/core/engine.py
U8NWXD: Yes, I kept this since I got errors without it

vivarium/core/engine.py
U8NWXD: That would be better! I'll fix it

vivarium/core/process.py
U8NWXD: nope! I'll remove it

vivarium/core/engine.py
eagmon: Let's keep the backwards compatibility of derivers as simple as possible. I think just pull all steps/derivers out of `processes` and put them in `self.step_paths`. If they are not included in `flow`, then assume deriver-like behavior. We can take the last of these deriver-steps, and make the first flow-steps depend on it -- or make them depend on all deriver-steps. Alternatively, don't even put dependence between deriver-steps and flow-steps, and we'll deal with issues if they come up.

vivarium/core/engine.py
U8NWXD: I added a test and fixed the issues that the test found

vivarium/core/engine.py
U8NWXD: done. I'm not holding the derivers in a separate list in `StepGraph` so that it's easy to return those as the first layers

vivarium/core/engine.py
U8NWXD: done

vivarium/core/process.py
U8NWXD: done

U8NWXD:
@eagmon I've addressed your comments. Could you take another look please?

eagmon (approved):
This looks great! Thanks @U8NWXD

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.4.0
+
+* Replaces `Deriver`s with `Step`s. While derivers were executed sequentially, steps are executed in topological generations according to a dependency graph. This lets some derivers run in parallel. This change mostly preserves backwards-compatibility since `Deriver` is now an alias for `Step`, and we still support legacy derivers that are specified without dependencies. These legacy derivers are executed sequentially before any steps. However, the minor version is incremented because the following public interfaces have changed (though we don't expect this to break dependent code):
+
+  * `Composite`s now have 2 more keys: `steps` and `flow`.
+  * `Engine.run_derivers()` has been replaced by `Engine.run_steps()`.
+  * `Engine.deriver_paths` has been replaced by `Engine.step_paths`.
```
U8NWXD, line 9: make these private

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  * New lists of step and flow updates have been added to the tuple returned by `Store.apply_update()`.
+  * `Store.EMPTY_UPDATES` has two more `None` values.
```
U8NWXD, line 11: Make this private

U8NWXD:
I added a changelog and made some more attributes private. I'm trying to be more aggressive about making attributes private so that we don't have as many breaking public API changes in the future

PR #103, opened 2021-11-03
Fix Store.generate() to let initial state override _value in schemas

U8NWXD:
In Engine, we use `vivarium.core.store.generate_state()` to generate `Engine.state`. In vivarium-core==0.3.14, `store.generate_state()` worked like this:

```python
    store = Store({})
    store.generate_paths(processes, topology)
    store.apply_subschemas()
    store.set_value(initial_state)
    store.apply_defaults()
    return store
```

Notice that it applies the subschemas before setting values using `initial_state`.

In vivarium-core==0.4.0, I wanted to make `Store.generate_paths()` private, so I did this instead for `store.generate_state():`

```python
    store = Store({})
    steps = steps or {}
    store.generate(tuple(), processes, steps, flow, topology, initial_state)
    return store
```

I thought that this would work the same way as before, but `Store.generate()` looks like this:

```python
        target = self.establish_path(path, {})
        target._generate_paths(processes, flow, topology)
        target._generate_paths(steps, flow, topology)
        target.generate_value(initial_state)
        target.apply_subschemas()
        target.apply_defaults()
```

Notice that here, we call `generate_value()` before `apply_subschemas()`, which causes errors to be thrown when the initial state conflicts with the schema. 

This PR restores the previous behavior by changing `Store.generate()` to swap `generate_value()` and `apply_subschemas()`. Now anything in initial_state will override anything in the schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.4.1
+
+* Fix a bug in `Store.generate()` that caused conflicts between a user-provided initial state and a schema to raise an error instead of the initial state taking priority.
+
 ## v0.4.0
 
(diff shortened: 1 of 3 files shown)


eagmon (approved):
good find! This ordering makes sense.

PR #110, opened 2021-11-30
Initial state is not modified when dividing cell.

matt-wolff:
When applying the divided values of both daughter cells to the default `initial_state` to get the new initial states, we were modifying `initial_state` in place, making it so the daughters shared the same dictionaries in memory for their values even when they had different stores. To get around this, I used `copy.deepcopy(initial_state)` since `dict(initial_state)`, which is recommended by the `deep_merge` function definition, is only a shallow copy, and thus creates the same issue of shared dictionaries.


eagmon (approved):
Thanks @matt-wolff! I'll merge this in and make a release later on today -- I hope to get #109 merged in for that same release.

PR #113, opened 2021-12-07
Let Vivarium Engine Profile (including parallel processes)

U8NWXD:
If you set `profile=True` in the `Engine` constructor, then when you call `Engine.end()`, Vivarium will compute profile statistics and store them in `Engine.stats` for you to analyze. The Engine also profiles parallel processes and includes those stats in `Engine.stats`, giving you a full picture of what's taking up runtime in your simulation.

See `vivarium/experiments/test_profile.py` for an example of how to use this.

### DIFF
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -6,6 +6,8 @@ Engine
 Engine runs the simulation.
 """
 
+import cProfile
+import pstats
 import os
 import logging as log
 import pprint
@@ -372,6 +374,7 @@ class Engine:
             emit_step: float = 1,
             display_info: bool = True,
             progress_bar: bool = False,
+            profile: bool = False,
     ) -> None:
         """Defines simulations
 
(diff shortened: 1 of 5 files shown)


vivarium/core/engine.py
```diff
@@ -415,12 +418,19 @@ def __init__(
             progress_bar: shows a progress bar
             emit_config: If True, this will emit the serialized
                 processes, topology, and initial state.
+            profile: Whether to profile the simulation with cProfile.
         """
+        if profile:
+            self.profiler = cProfile.Profile()
+            self.profiler.enable()
```
eagmon, line 425: You need `self.profiler = None` perhaps if `profile=False`

vivarium/experiments/test_profiler.py
```diff
@@ -0,0 +1,59 @@
...
+        profile=True,
+    )
+    engine.update(3)
+    engine.end()
+    assert engine.stats is not None
+    stats = engine.stats.strip_dirs()
+    process_a_runtime = stats.stats[  # type: ignore
+        ('test_profiler.py', 16, 'next_update')][3]
+    process_b_runtime = stats.stats[  # type: ignore
+        ('test_profiler.py', 31, 'next_update')][3]
```
eagmon, line 56: Out of curiosity, what are those `16`, `31` for?

U8NWXD: Those are the line numbers for where the two `next_update` functions are defined. That's how `cProfile` resolves cases like this where there are 2 functions of the same name in a file

vivarium/core/engine.py
eagmon: looks like you took car of this.

U8NWXD: I fixed this when fixing some failing tests

eagmon (approved):
This is great! I'm tempted to remake `profile_runtime.py` with this.

U8NWXD:
Thanks @eagmon! I'll merge it and create a new release

PR #116, opened 2021-12-08
expire and recache topology_views after Engine.apply_update

eagmon:
`Store.build_topology_views` allows Engine to expire and recache views upon a hierarchy update. This worked great from simulations such as `vivarium-ecoli`, which have infrequent hierarchy updates, but it leads to considerable slowdown for division-heavy simulation such as `tumor-tcell`. This PR attempts to improve the runtime for `tumor-tcell` found by @johnhickey22.

This PR introduces a new `test_hyperdivision` function to `engine_tests.py` which reproduced the high-division simulations and consequent slowdown. This was fixed by removing a line with `state = self` in `Store.schema_topology`, which was returning the entire agents branch for `division` processes that link to agents with the topology `{'agents_port': ('..', '..', 'agents')}`. This would result an enormous slowdown when `view_values` was called on that view of agents.

Additionally, this PR makes it so that schema with `{ 'agents': {'*': {}}` sees the top-level IDs under that port, and `{ 'agents': {}` does not see anything, but can be used for output (especially useful for division updates).

### DIFF
--- a/vivarium/composites/toys.py
+++ b/vivarium/composites/toys.py
@@ -833,9 +833,7 @@ class ToyDividerProcess(Process):
 
     def next_update(self, timestep, states):
         x = states['variable']['x']
-        # print(x, self.parameters['x_division_threshold'])
         if x > self.parameters['x_division_threshold']:
-            # print('DIVIDE!')
             daughter_ids = [
                 str(self.agent_id) + '0',
                 str(self.agent_id) + '1']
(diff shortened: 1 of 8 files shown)


eagmon:
Note that to this point the changes have not significantly improved `tumor-tcell's` performance. My guess is that the slowdown has to do with too frequent topology re-caching resulting from frequent divisions. I'll make a test to try and reproduce this.

eagmon:
@johnhickey22 confirms this fixed the runtime on high division experiments! Merging it in now.

PR #119, opened 2021-12-13
Replace isinstance(..., Step) with Process.is_step() calls

U8NWXD:
Instead of using `isinstance` to check whether a process is a step, we should use `Process.is_step()` since some processes don't inherit from `Step` but still override `is_step()` to return `True`.

This PR also adds an assertion to make sure that when Engine treats a process like a step, that process really is a step.


eagmon (approved):
looks good!

PR #121, opened 2021-12-17
Use initial instead of mother state for division

U8NWXD:
Division used the mother cell state as the default state and deep-merged the result of running the dividers into that default. This led to daughter cells getting molecules that were sent by the dividers to the other daughter.

Instead, we should use the result of running the dividers as the default daughter state. Then, we can deep-merge in any state provided by the meta-division process as part of the `_divide` update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.4.14
+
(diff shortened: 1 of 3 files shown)


eagmon (approved):

prismofeverything:
Good catch @U8NWXD! Looks good to me.

prismofeverything (approved):

PR #126, opened 2021-12-28
Engine.run_for allows iterative engine calls without completing.

eagmon:
In order to update Engine's processes iteratively we have to allow processes to hang on the `front`, and not be forced to update within the given interval. To do this, I separated `Engine.update` from a new method `Engine.run_for`. `Engine.update` completes all the processes within the provided interval, just as it had before. `Engine.run_for` can be called iteratively to advance the front up to the interval, but can leave processes hanging if its argument `force_complete=False`. The new method `Engine.complete()` forces all processes to complete at the current `global_time`.

Note:
 - To get an `EngineProcess` that can handle multi-timestepping will require its `Process.calculate_timestep` to call a currently non-existent `Engine.next_timestep` method, which will return the time to the next process's update.

### DIFF
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -878,6 +878,9 @@ class Engine:
             self.print_summary(clock_finish)
 
     def complete(self) -> None:
+        """
+        Force all processes on the front to complete at the current global time
+        """
         self.run_for(interval=0, force_complete=True)
 
     def run_for(
@@ -954,7 +957,7 @@ class Engine:
                             self.front[path]['update'] = update
 
                         else:
-                            # mark this path as "quiet" so its time can be advanced
+                            # mark this path "quiet" so its time can be advanced
                             self.front[path]['update'] = (EmptyDefer(), store)
                             quiet_paths.append(path)
 
@@ -985,7 +988,8 @@ class Engine:
                 updates = []
                 paths = []
                 for path, advance in self.front.items():
-                    if advance['time'] <= self.global_time and advance['update']:
+                    if advance['time'] <= self.global_time \
+                            and advance['update']:
                         new_update = advance['update']
                         updates.append(new_update)
                         advance['update'] = {}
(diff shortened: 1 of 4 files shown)


vivarium/core/engine.py
```diff
@@ -850,22 +860,48 @@ def update(
...
         """
-
-        time = 0.0
-        emit_time = self.emit_step
+        Run each process for the given interval and update the states.
+        """
         clock_start = clock.time()
+        self.run_for(interval=interval, force_complete=True)
+        self.check_complete()
+        clock_finish = clock.time() - clock_start
```
U8NWXD, line 869: To me, `start` and `finish` usually mean absolute times, so `finish - start = delta`. Here, `clock_finish` is more of what I would think of as the delta

vivarium/core/engine.py
```diff
@@ -850,22 +860,48 @@ def update(
...
+            interval: the amount of time to simulate the composite.
+            force_complete: a bool indicating whether to force processes
+                to complete at the end of the interval.
+        """
+        end_time = self.global_time + interval
+        emit_time = self.global_time + self.emit_step
+        run_complete = False
 
-        while time < interval:
+        while not run_complete:
```
U8NWXD, line 904: Why not just make this condition `self.global_time == end_time`? I think that would be clearer

eagmon: @U8NWXD -- This while loops needs to actually run at `self.global_time == end_time`. So `run_complete` is set to True only after all the processes that need to complete at `end_time` are updated.

U8NWXD: Then couldn't you make it `self.global_time < end_time`? Or is it possible for `self.global_time` to not be increased?

eagmon:
@U8NWXD -- I improved the `run_for` while loop logic a bit. The problem with `self.global_time < end_time` is that I still wanted to retain the ability to force processes to complete at `end_time`, in order to reproduce the previous `Engine.update` behavior.  Now I use `while self.global_time < end_time or force_complete`, which allows the processes to complete with the `force_complete` flag.

PR #127, opened 2022-01-05
Store API tests which fail

WilliamIX:
This PR creates two new tests in store_api.py which currently fail. 

The first is `test_run_inserted_store` which produces an error `store at path ('p1',) does not have a topology_view` by creating a store and manually inserting a process using the API.

The second is similar to the first, but two processes are rewired after being created using store.connect. This results in a different error: `'ToyProcess' object is not subscriptable`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v0.4.17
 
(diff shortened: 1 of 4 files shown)


eagmon (approved):
Thanks @WilliamIX! simple fix for all of these -- calling `store.build_topology_view()` in the `Engine` constructor.

PR #131, opened 2022-01-12
set_path with inserted process does not use process.initial_state()

eagmon:
This corrects a bug found by @U8NWXD -- inserting a process with the `Store` api should not assume that the process's initial_state is used.

### DIFF
--- a/vivarium/core/store.py
+++ b/vivarium/core/store.py
@@ -362,7 +362,10 @@ class Store:
                 self.insert({
                     'processes': value.generate_processes({'name': final}),
                     'topology': value.generate_topology({'name': final}),
-                    'initial_state': value.initial_state()})
+                    # If a process's initial_state is expected, this will need to use inverse_topology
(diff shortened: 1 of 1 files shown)


vivarium/core/store.py
```diff
@@ -362,7 +362,10 @@ def set_path(self, path, value):
                 self.insert({
                     'processes': value.generate_processes({'name': final}),
                     'topology': value.generate_topology({'name': final}),
-                    'initial_state': value.initial_state()})
+                    # If a process's initial_state is expected, this will need to use inverse_topology
+                    # 'initial_state': value.initial_state()
```
U8NWXD, line 366: Can we just remove this instead? We have the git history, so I don't see any benefit to leaving the commented-out code

eagmon: sure.

U8NWXD (approved):

PR #132, opened 2022-01-13
Remove Reference to Leaf From No Path Error Msg

U8NWXD:
Consider the following tree:

```
          a
         /  \
       b     c
      /  \     \
     d    e     f
```

The path `('e',)` from node `a` is invalid, but there is a path from `a` to `e` and `a` is not a leaf node. This PR corrects Store's error message to accurately describe cases like this.


eagmon (approved):
yes!

PR #135, opened 2022-01-24
Run CI checks on Python 3.9 and Python 3.10

U8NWXD:
Expand supported Python versions to include versions 3.9 and 3.10


1fish2 (approved):
LGTM.

I assume you'll push the `install_requires` `numpy>=1.22.1` and `scipy>=1.7.3` through to vivarium-ecoli. Or is it too aggressive to push these install_requires on all vivarium clients?

Since Python 3.10 deprecates distutils, we should plan to replace distutils in all vivarium projects. setuptools embeds a forked (and updated?) copy of distutils but there are compatibility issues. It's a mess.

(Presumably wcEcoli will stay on Python 3.8.7 and distutils. I did try updating it to setuptools but reverted that due to https://github.com/CovertLab/wcEcoli/issues/1113.)

`vivarium-core` should consider using something more modern like [Flit](https://flit.readthedocs.io/en/latest/index.html). Experts recommend Flit, and that should be easier and more robust to use and maintain. Flit handles the normal cases with a small, declarative `pyproject.toml` file instead of `setup.py`. `setup.py` is not dead but apparently we [should not run it directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

Flit does not support Cython code so it won't work for `vivarium-ecoli`. Given that, maybe we should learn and use a different tool everywhere like maybe [Pipenv](https://pipenv.pypa.io/en/latest/) or [Bento](https://cournape.github.io/Bento/) or something. AIUI, Numpy & Scipy are moving to [Meson](https://mesonbuild.com/index.html) which is a build tool, not a packaging tool. Maybe it includes Python packaging features. Anyway, don't switch to `numpy.distutils`.

setup.py
```diff
@@ -52,14 +52,14 @@
         long_description_content_type='text/markdown',
         package_data={},
         include_package_data=True,
-        python_requires='>=3.6, <4',
+        python_requires='>=3.8.*, <=3.10.*',
```
U8NWXD, line 55:
```suggestion
        python_requires='>=3.8, <3.11',
```

PR #147, opened 2022-01-28
Pass Parameters Positionally in __setstate__

U8NWXD:
If a user creates a process whose constructor accepts parameters under a
customized name (e.g. `parameters` instead of `initial_parameters`),
serialization with `__setstate__` should still work. To support this,
make `__setstate__` pass parameters to the constructor as a positional,
not a keyword, argument.

repository.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -151,9 +151,9 @@ class Process(metaclass=abc.ABCMeta):
                 'not supported.')
         return self._original_parameters
 
-    def __setstate__(self, state: dict) -> None:
+    def __setstate__(self, parameters: dict) -> None:
         """Initialize process with parameters"""
-        self.__init__(state)  # type: ignore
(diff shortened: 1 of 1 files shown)


eagmon (approved):

vivarium/core/process.py
```diff
@@ -153,7 +153,7 @@ def __getstate__(self) -> dict:
 
     def __setstate__(self, state: dict) -> None:
         """Initialize process with parameters"""
-        self.__init__(parameters=state)  # type: ignore
+        self.__init__(state)  # type: ignore
```
eagmon, line 156: can you call this `parameters`, instead of `state`?

eagmon (changes requested):

vivarium/core/process.py
U8NWXD: done!

Issue #146, opened 2022-01-28
Update Documentation


U8NWXD:
Issues to fix:

* [x] Python version requirement is now 3.8-3.10, not 3.6
  * https://vivarium-core.readthedocs.io/en/latest/getting_started.html#python-3
  * https://vivarium-core.readthedocs.io/en/latest/getting_started_dev.html#python-3
* [x] Verify that getting started instructions are still valid: https://vivarium-core.readthedocs.io/en/latest/getting_started.html
* [x] `CONTRIBUTING.md` should point to the getting started guide
* [x] Verify that the `vivarium-template` repository is still correct vivarium-collective/vivarium-template#1
* [x] Replace "compartment" with "composite" or "composer"
  * https://vivarium-core.readthedocs.io/en/latest/getting_started.html#repository-structure
  * https://vivarium-core.readthedocs.io/en/latest/getting_started.html#some-terminology-processes-and-compartments
  * https://vivarium-core.readthedocs.io/en/latest/guides/compartments.html
  * https://vivarium-core.readthedocs.io/en/latest/guides/processes.html
  * https://vivarium-core.readthedocs.io/en/latest/tutorials/write_process.html
* [x] The getting started guide for developers doesn't really cover how to develop: https://vivarium-core.readthedocs.io/en/latest/getting_started_dev.html
* [ ] Verify that the notebook tutorials still work
* [x] The "modeling paradigm" guide frames Vivarium as ABM + WCM. Is this still appropriate? https://vivarium-core.readthedocs.io/en/latest/guides/modeling_paradigm.html
* [x] Update the [processes guide](https://vivarium-core.readthedocs.io/en/latest/guides/processes.html):
  * Ports are no longer defined in constructor
  * We can now just call `super()`
  * Ports can be wired directly to variables
  * Derivers are now steps
  * `is_deriver()` and `is_step()` methods
  * Custom timestep
  * Process names
  * Conditional updating
* [x] Discuss flows in compartments guide: https://vivarium-core.readthedocs.io/en/latest/guides/compartments.html
* [ ] Document Composer and MetaComposer better
* [ ] Document the available arguments for the Engine constructor
* [x] Document the `**` syntax: https://vivarium-core.readthedocs.io/en/latest/guides/hierarchy.html
* [x] `ConveneinceKinetics` is now it's own project: https://vivarium-core.readthedocs.io/en/latest/tutorials/write_process.html
[...]

PR #150, opened 2022-02-02
Update docs

### LINKED ISSUE #101: Decide on Policy for Breaking API Changes
Before we release v1.0.0, we should document a policy for breaking API changes and versioning. I recommend we follow [semantic versioning](https://semver.org/) (which we mostly have been already) with the following further policies:

* Our supported API consists of all public, documented interfaces that are not marked as being experimental in their docstrings.
* Changes to the supported API will be reflected in the versioning according to semantic versioning. They will also be marked as deprecated in the documentation (and a warning will be raised if possible) for at least one minor release before they are removed.
* Experimental interfaces may be changed or removed at any time and in any version.
* The following are not considered breaking API changes:

  * Adding to a function a parameter with a default value (i.e. an optional parameter)
  * Adding a new key to a dictionary. For example, expanding the `Composite` dictionary with an extra key.
  * When a function accepts a dictionary as an argument, adding more, optional keys to that dictionary. For example, letting the user specify new configuration options through a `config` dictionary.

U8NWXD:
Update documentation for v1.0.0.

Fixes #101.
Fixes part of #146.

repository.


eagmon (approved):
Thanks Chris!

PR #157, opened 2022-02-08
Composite.generate_store flows and steps

eagmon:
adds flows and steps to `Composite.generate_store`

repository.


U8NWXD:
Do you mean `generate_state` instead of `generate_store`?

eagmon:
This is `Composite. generate_store()`, which calls `generate_state`. `generate_state` already had flows and steps, just had to pass them in with `Composite. generate_store()`

PR #160, opened 2022-02-08
Support serialization with original parameters from subclasses

### LINKED ISSUE #159: Process Serialization with Inheritance
Currently, `Process.__init__()` saves off a copy of the parameters to `self._original_parameters`. Then when serializing, `Process.__getstate__()` returns `self._original_parameters` and `Process.__setstate__()` calls `self.__init__()` with the parameters.

This is problematic in class hierarchies where constructors change parameters. For example:

```python
class A(Process):

    def __init__(self, parameters):
        super().__init__({'2': parameters['1']})

a = A({'1': True})
a2 = pickle.loads(pickle.dumps(a))
```

Here, `A` receives a parameter under the key `1` but passes that parameter to `Process` under key `2`. 

Here's how the serialization/deserialization flow looks:

1. When an instance of `A` is serialized, `Process.__getstate__()` returns `{'2': True}`.
2. Then when we want to deserialize and get `a2`, `Process.__setstate__()` calls `self.__init__(parameters)`. However, it is `A.__init__()` that runs here, not `Process.__init__()`, so `A.__init__()` will be looking for a nonexistent key `1` in the parameters.

The fact that `A.__init__()` runs is correct. There could be code there that we need to run to initialize instance variables. The problem is that `a.__getstate__()` should be returning the original parameters passed to `A.__init__()`, not those passed to `Process.__init__()`.

Here are some possible solutions:

1. Add an optional parameter to `Process.__init__()` for the original subclass parameters, which will then be saved in `self._original_parameters`. This route will require that we update all constructors in an inheritance hierarchy to handle the new parameter.
2. Add a new `Process.save_parameters()` method that subclasses can call to save off their original parameters. With this approach, we will only have to change `Process` and the class at the bottom of the hierarchy. However, we will have to handle multiple calls to `Process.save_parameters()` since we only want to save the parameters from the child-most subclass.
3. Add a new special key to the `parameters` dictionary for the original parameters. If provided, `Process.__init__()` will save this in `self._original_parameters` instead of `parameters`.

I prefer approach (3) since I think it's most in line with how we pass other special configs to `Process`.

U8NWXD:
Fixes #159. This PR will also be the v1.0.1 release.

repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v1.0.1
+
+* (#151) Make `Engine._run_steps()` public.
+* (#155) Add error messages for flows that include steps not found in
+  `Engine.steps` and for dividers specified as strings not found in the
+  divider registry.
+* (#157) Add flows and steps to the value returned by
+  `Composite.generate_store()`.
+
 ## v1.0.0
 
 * (#131) Do not assume that a process's `initial_state` should be used
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import re
 from setuptools import setup
 
 
-VERSION = '1.0.0'
+VERSION = '1.0.1'
 
 
 if __name__ == '__main__':
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -98,9 +98,20 @@ class Process(metaclass=abc.ABCMeta):
         work properly.
 
         Args:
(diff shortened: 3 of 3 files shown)


eagmon (changes requested):
Just one request -- can you document this special key in the docstring?

U8NWXD:
@eagmon I've updated the docstring

eagmon (approved):

PR #161, opened 2022-02-11
initial_state added to Composite

eagmon:
Initial_state was always handled separately from composites, which are dicts of processes, steps, topology, and flow. This PR adds `state` to `Composite`, which makes it sufficient to specify a full store hierarchy. 

This PR also adds the function arguments `processes`, `topology`, `steps`, `flow`, `state`, and `store` to `Composite` -- previously it only accepted a `config` dict with these arguments as keys. Function arguments are just a cleaner way of doing the same thing. `config` dicts are still compatible as the 1st positional argument to `Composite`.

repository.

### DIFF
--- a/vivarium/core/composer.py
+++ b/vivarium/core/composer.py
@@ -158,7 +158,7 @@ class Composite(Datum):
         """
         config = config or {}
         initial_state = config.get('initial_state', {})
-        initial_state = deep_merge(self.state, initial_state)
+        initial_state = deep_merge(copy.deepcopy(self.state), initial_state)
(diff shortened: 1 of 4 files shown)


prismofeverything (approved):
Excellent! Was just going to suggest this then saw you already had a PR up for it.

prismofeverything:
In a related topic, which branch is your Composite API in? I just wanted to take a look at what you were thinking there.

eagmon:
@prismofeverything -- https://github.com/vivarium-collective/vivarium-core/tree/store-composite

PR #167, opened 2022-02-23
Make Serializers Robust and Structured

U8NWXD:
Make our handling of serialization more structured and robust. Now, we create classes that inherit from `Serializer` to define serializers. These serializers each know what kinds of objects they can serialize and deserialize, so the `serialize_value` and `deserialize_value` functions can automatically find the correct serializer for a given object. We also throw an error if no serializer or multiple serializers are found for an object to make serialization more robust.

repository.

### DIFF
--- a/vivarium/core/serialize.py
+++ b/vivarium/core/serialize.py
@@ -56,10 +56,10 @@ class IdentitySerializer(Serializer):
     '''Serializer for base types that get serialized as themselves.'''
 
     def can_serialize(self, data):
-        if isinstance(data, (int, float, bool)):
+        if (
+                isinstance(data, (int, float, bool, str))
+                and not NumpyScalarSerializer.can_serialize(data)):
             return True
-        if isinstance(data, str):
-            return not self.REGEX_FOR_SERIALIZED_ANY_TYPE.fullmatch(data)
         if data is None:
             return True
         return False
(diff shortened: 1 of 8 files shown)


vivarium/core/serialize.py
```diff
@@ -1,21 +1,417 @@
...
+class NumpySerializer(Serializer):
+    """Serializer for Numpy arrays.
+
+    Numpy array serialization is lossy--we serialize to Python lists, so
+    deserialization will produce a Python list instead of a Numpy array.
+    """
+
+    def __init__(self) -> None:
+        super().__init__(exclusive_types=(np.ndarray,))
+
```
eagmon, line 104: minor -- there's an extra space between these methods.

vivarium/core/serialize.py
```diff
@@ -1,21 +1,417 @@
...
+        raise NotImplementedError(
+            'ObjectIds cannot be deserialized.')
 
 
 def composite_specification(
         composite: Composite,
         initial_state: bool = False,
 ) -> dict:
     """Return the serialized specification for a :term:`composite` instance"""
+    warnings.warn(
```
eagmon, line 412: Is this equivalent to serializing a composite? The reason for this method was to provide an entirely data representation of a composite, which could ideally be re-initialized again. I was thinking of moving it to `composer.py`.

eagmon (approved):
This is a great update! I have just a few minor requests/questions.

vivarium/core/serialize.py
eagmon: I would like it back because I know there are a few cases of it out there. Maybe just have it call `serialize_value` and also give a depreciation warning? I don't remember why the `_multi_update` key sneaked in there, it really doesn't make sense but I know it showed up. `_schema` is an unfortunate default key of composites.

U8NWXD: I removed the deprecation warning to keep `composite_specification` as-is

PR #169, opened 2022-03-14
`get_history_data_db` remove time key requirement

eagmon:
Some unassembled data might not have a 'time' variable -- this includes some large 'unique' states in vivarium-ecoli that need to get merged into a dict with ['bulk', 'listeners', 'time'] keys, which has time. Removing this requirement for a 'time' key from `get_history_data_db` recovers some lost 'unique' states.

repository.

### DIFF
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -422,7 +422,7 @@ def get_history_data_db(
     cursor = history_collection.find(experiment_query, projection)
     raw_data = []
     for document in cursor:
-        if document['data'].get('time'):
+        if document.get('assembly_id'):
             raw_data.append(document)
 
     # re-assemble data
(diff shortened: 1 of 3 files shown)


vivarium/core/emitter.py
```diff
@@ -422,7 +422,7 @@ def get_history_data_db(
     cursor = history_collection.find(experiment_query, projection)
     raw_data = []
     for document in cursor:
-        if document['data'].get('time'):
+        if document.get('assembly_id'):
```
U8NWXD, line 425: It doesn't look like we'll every have something in the history collection without an assembly ID:

https://github.com/vivarium-collective/vivarium-core/blob/c0966b8057221ccd8ce2c47d124b1bce240f2541/vivarium/core/emitter.py#L322-L334

Therefore, what if we asserted the presence of `assembly_id` instead? That way, if we ever run into a bug like this again, we'll get an assertion error instead of a silent failure

U8NWXD (changes requested):

vivarium/core/emitter.py
eagmon: done!

PR #170, opened 2022-03-14
Toggle emits on/off in Store

eagmon:
Previously it was necessary to override variable emits with a '_schema' value passed to the processes or the composite. This was confusing and complicated because it had to be directed through the processes' ports, rather than through the Store hierarchy.

`Store.toggle_emits` allows you to turn emits on/off with more ease. This method has args `emit` and `paths`. `emit` is a boolean True/False to turn the emit on or off. `paths` is a list of paths to toggle on/off -- all nodes below those paths will be set to the `emit` value.

This can be accessed through `Engine` with the `store_emit` keyword, which can be supplied with a dictionary that has `'on'`, `'off'` keys that map to a list of paths to turn on or off: 

```
sim= Engine(
    composite=composite,
    store_emit={
        'on': [('path', '1',), ('path', '2',)],
    }
)
``` 

This allows you to turn off all emits like this:

```
sim= Engine(
    composite=composite,
    store_emit={
        'off': [()],
    }
)
``` 

repository.

### DIFF
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -424,9 +424,12 @@ class Engine:
                 provide as the value for the key ``experiment_id``.
             display_info: prints experiment info
             progress_bar: shows a progress bar
-            store_emit: An optional dictionary to turn on/off emits. This
-                dictionary requires the keys `on`/`off` mapping to a list
-                of paths in the Store hierarchy to be toggled on or off.
+            store_emit: An optional dictionary to turn emits on or off. This
+                dictionary requires the keys (`on`,`off`), mapping to a list
+                of paths in the Store hierarchy to be turned on or off. The
+                on configs take precedence over the off configs in that all
+                paths in `off` are turn off first, and can be turned on again
+                by the paths in `on`.
             emit_topology: If True, this will emit the topology with the
                 configuration data.
             emit_processes: If True, this will emit the serialized
@@ -482,8 +485,10 @@ class Engine:
 
         # override emit settings in store
         if store_emit:
-            self.state.toggle_emits(emit=False, paths=store_emit.get('off', []))
-            self.state.toggle_emits(emit=True, paths=store_emit.get('on', []))
+            self.state.set_emit_values(
+                paths=store_emit.get('off', []), emit=False)
+            self.state.set_emit_values(
+                paths=store_emit.get('on', []), emit=True)
 
         # settings for self._emit_configuration()
         self.emit_topology = emit_topology
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -482,12 +482,8 @@ class Engine:
 
         # override emit settings in store
         if store_emit:
-            turn_off = store_emit.get('off')
-            turn_on = store_emit.get('on')
-            if turn_off:
-                self.state.toggle_emits(emit=False, paths=turn_off)
-            if turn_on:
-                self.state.toggle_emits(emit=True, paths=turn_on)
+            self.state.toggle_emits(emit=False, paths=store_emit.get('off', []))
+            self.state.toggle_emits(emit=True, paths=store_emit.get('on', []))
 
         # settings for self._emit_configuration()
         self.emit_topology = emit_topology
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -425,7 +425,7 @@ class Engine:
             display_info: prints experiment info
             progress_bar: shows a progress bar
             store_emit: An optional dictionary to turn emits on or off. This
-                dictionary requires the keys (`on`,`off`), mapping to a list
+                dictionary may contain the keys (`on`,`off`), mapping to a list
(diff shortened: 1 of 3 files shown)


vivarium/core/engine.py
```diff
@@ -423,8 +424,15 @@ def __init__(
                 provide as the value for the key ``experiment_id``.
             display_info: prints experiment info
             progress_bar: shows a progress bar
-            emit_config: If True, this will emit the serialized
-                processes, topology, and initial state.
+            store_emit: An optional dictionary to turn on/off emits. This
+                dictionary requires the keys `on`/`off` mapping to a list
+                of paths in the Store hierarchy to be toggled on or off.
```
U8NWXD, line 429: Specify that the `on` configs take precedence over the `off` configs?

vivarium/core/engine.py
```diff
@@ -472,6 +480,16 @@ def __init__(
...
         self.emitter: Emitter = get_emitter(emitter_config)
 
+        # override emit settings in store
+        if store_emit:
+            turn_off = store_emit.get('off')
+            turn_on = store_emit.get('on')
+            if turn_off:
+                self.state.toggle_emits(emit=False, paths=turn_off)
+            if turn_on:
+                self.state.toggle_emits(emit=True, paths=turn_on)
```
U8NWXD, line 490: Possibly cleaner:

```suggestion
            self.state.toggle_emits(emit=False, paths=store_emit.get('off', []))
            self.state.toggle_emits(emit=True, paths=store_emit.get('on', []))
```

vivarium/core/store.py
```diff
@@ -1006,6 +1006,29 @@ def emit_data(self):
             return self.value
         return None
 
+    def toggle_emits(self, emit=False, paths=None):
```
U8NWXD, line 1009: To me, "toggle" means to convert on to off and off to on. Instead, this function sets emits to be on/off. For example if you call `toggle_emits(emit=True, paths=[('a',)])` and `('a',)` already has its emit enabled, this function doesn't toggle the emit to off.

vivarium/core/store.py
```diff
@@ -1006,6 +1006,29 @@ def emit_data(self):
...
+        """
+        if path:
+            assert isinstance(path, tuple), 'path must be a tuple'
+            target = self.get_path(path)
+            target.toggle_emit(emit=emit)
+        elif self.inner:
+            for child in self.inner.values():
+                child.toggle_emit(emit=emit)
+        else:
+            self.emit = emit
```
U8NWXD, line 1030: Can branch nodes have an emit setting? I don't think we want to allow that, but I'm not sure we actually forbid it

vivarium/core/store.py
```diff
@@ -1006,6 +1006,29 @@ def emit_data(self):
             return self.value
         return None
 
+    def toggle_emits(self, emit=False, paths=None):
+        """
+        Turn on/off emits for all inner nodes of the list of paths.
+        """
+        if paths:
+            assert isinstance(paths, list), 'paths must be a list'
```
U8NWXD, line 1014: I think the only real requirement is that `paths` be iterable, which is already going to be enforced by the `for` loop below

U8NWXD (changes requested):

vivarium/core/store.py
eagmon: I renamed these methods `set_emit_values` and `set_emit_value`

vivarium/core/store.py
eagmon: It looks like we don't support branch node emits: https://github.com/vivarium-collective/vivarium-core/blob/7af1eb0593ff05f9fb08ce3542f29585f7866307/vivarium/core/store.py#L732-L747

eagmon: Although this might allow it... https://github.com/vivarium-collective/vivarium-core/blob/7af1eb0593ff05f9fb08ce3542f29585f7866307/vivarium/core/store.py#L652

eagmon: And here as well, only for leaf nodes: https://github.com/vivarium-collective/vivarium-core/blob/7af1eb0593ff05f9fb08ce3542f29585f7866307/vivarium/core/store.py#L652

vivarium/core/engine.py
```diff
@@ -423,8 +424,18 @@ def __init__(
                 provide as the value for the key ``experiment_id``.
             display_info: prints experiment info
             progress_bar: shows a progress bar
-            emit_config: If True, this will emit the serialized
-                processes, topology, and initial state.
+            store_emit: An optional dictionary to turn emits on or off. This
+                dictionary requires the keys (`on`,`off`), mapping to a list
```
U8NWXD, line 428:
```suggestion
            store_emit: An optional dictionary to turn emits on or off. This
                dictionary may contain the keys `on` and `off`, mapping to a list
```

Each key is optional, right? You could provide an empty dict, in which case nothing would happen

U8NWXD (changes requested):
just one last nit

PR #175, opened 2022-03-23
override schema directly in Engine

eagmon:
A new `store_schema` arg to `Engine` allows us to override schema directly. This includes state values, emits, etc. This replaces the `store_emit` argument introduced in #170, since the new `store_schema` dict can turn emits on/off more directly. A new check in `Store._apply_config` lets us use schema to set the emit values for all leaf nodes under a given branch, by setting the `_emit` schema of that branch node to `True` or `False`.

In the new `test_add_new_state`, this feature is demonstrated by adding an `extra` key to the agent state, and turning its emit to true:
```
    new_schema = {'agents': {agent_id: {'extra': {'_emit': True, '_value': 1.0}}}}
    experiment = Engine(
        processes=composite.processes,
        topology=composite.topology,
        store_schema=new_schema,
    )
```

As an example of how to set branch emit values, here we set the entire tree to emit by setting the top-level emit value to True:
```
    exp = Engine(
        composite=composite,
        store_schema={'_emit': True}
    )
```

TODO:
 - [X] document `store_schema`
 - [x] expand schema enforcement of updaters
 
repository.

### DIFF
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -371,6 +371,7 @@ class Engine:
             description: str = '',
             emitter: Union[str, dict] = 'timeseries',
             store_emit: Optional[dict] = None,
+            store_schema: Optional[dict] = None,
             emit_topology: bool = True,
             emit_processes: bool = False,
             emit_config: bool = False,
@@ -484,6 +485,8 @@ class Engine:
         self.emitter: Emitter = get_emitter(emitter_config)
 
         # override emit settings in store
+        if store_schema:
+            self.state._apply_config(store_schema)
         if store_emit:
             self.state.set_emit_values(
                 paths=store_emit.get('off', []), emit=False)
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -370,8 +370,8 @@ class Engine:
             metadata: Optional[dict] = None,
             description: str = '',
             emitter: Union[str, dict] = 'timeseries',
-            store_emit: Optional[dict] = None,
             store_schema: Optional[dict] = None,
+            store_emit: Optional[dict] = None,
             emit_topology: bool = True,
             emit_processes: bool = False,
             emit_config: bool = False,
@@ -425,6 +425,10 @@ class Engine:
                 provide as the value for the key ``experiment_id``.
             display_info: prints experiment info
             progress_bar: shows a progress bar
+            store_schema: An optional dictionary to expand the store hierarchy
+                configuration. This includes adding new values and turning emits
+                on. The dictionary needs to be structured as a hierarchy, which will
+                expand the existing store hierarchy.
             store_emit: An optional dictionary to turn emits on or off. This
                 dictionary may contain the keys (`on`,`off`), mapping to a list
                 of paths in the Store hierarchy to be turned on or off. The
--- a/vivarium/core/engine.py
+++ b/vivarium/core/engine.py
@@ -426,15 +426,10 @@ class Engine:
             display_info: prints experiment info
             progress_bar: shows a progress bar
             store_schema: An optional dictionary to expand the store hierarchy
-                configuration. This includes adding new values and turning emits
-                on. The dictionary needs to be structured as a hierarchy, which will
-                expand the existing store hierarchy.
-            store_emit: An optional dictionary to turn emits on or off. This
-                dictionary may contain the keys (`on`,`off`), mapping to a list
-                of paths in the Store hierarchy to be turned on or off. The
-                on configs take precedence over the off configs in that all
-                paths in `off` are turn off first, and can be turned on again
-                by the paths in `on`.
+                configuration, and also to turn emits on or off. The dictionary
+                needs to be structured as a hierarchy, which will expand the
+                existing store hierarchy. Setting an emit value for a branch node
+                will set the emits of all the branch's leaves to that value.
             emit_topology: If True, this will emit the topology with the
                 configuration data.
             emit_processes: If True, this will emit the serialized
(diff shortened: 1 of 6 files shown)


U8NWXD:
@eagmon does this suffer from the same issue with changes being lost upon division as #170?

eagmon:
@U8NWXD -- yes, same issue. With division new schema are created by the composer and inserted into the hierarchy without regards to the existing schema.

vivarium/core/engine.py
```diff
@@ -424,12 +424,11 @@ def __init__(
...
-                dictionary may contain the keys (`on`,`off`), mapping to a list
-                of paths in the Store hierarchy to be turned on or off. The
-                on configs take precedence over the off configs in that all
-                paths in `off` are turn off first, and can be turned on again
-                by the paths in `on`.
+            store_schema: An optional dictionary to expand the store hierarchy
+                configuration, and also to turn emits on or off. The dictionary
+                needs to be structured as a hierarchy, which will expand the
+                existing store hierarchy. Setting an emit value for a branch
+                node will set the emits of all the leaves to that value.
```
U8NWXD, line 431: So technically I think this is a breaking change since we're removing the `store_emit` flag which has already been released. It's coming so close that I doubt anyone has started using `store_emit` yet, but I think we should try to avoid these kinds of flip-flops in the future so that we don't break our API too often

U8NWXD (approved):

U8NWXD:
@thalassemia heads up that if you started using the `store_emit` option, this PR will change that to `store_schema`

vivarium/core/store.py
```diff
@@ -476,24 +476,45 @@ def _check_default(self, new_default):
...
+                new_schema[schema_key])
+        if (
+                current_schema
+                and current_schema != DEFAULT_SCHEMA
+                and current_schema != new_schema):
+            raise ValueError(
+                f"incompatible {schema_key} schema assignment: {new_schema} "
+                f"at {self.path_for()}. "
+                f"schema {current_schema} is already assigned.")
+        return new_schema
```
eagmon, line 517: @U8NWXD -- I pulled this method out from the `_divider` config check, and plan to use it for `_updater` as well. What do you think?

vivarium/core/store.py
```diff
@@ -476,24 +476,45 @@ def _check_default(self, new_default):
...
+        current_schema = self.__dict__[schema_key]
+        if current_schema is not None and current_schema != new_schema:
+            raise ValueError(
+                f"incompatible {schema_key} schema assignment: {new_schema} "
+                f"at {self.path_for()}. "
+                f"schema {current_schema} is already assigned.")
+        return new_schema
+
+    def _check_schema_set_default(self, schema_key, new_schema, schema_registry):
+        current_schema = self.__dict__[schema_key]
```
U8NWXD, line 502: I think `getattr` is a more standard way to do this:

```suggestion
        current_schema = getattr(self, schema_key)
```

vivarium/core/store.py
```diff
@@ -476,24 +476,45 @@ def _check_default(self, new_default):
...
+        if isinstance(new_schema, str):
+            new_schema = schema_registry.access(new_schema)
+        if isinstance(new_schema, dict) and isinstance(
+                new_schema[schema_key], str):
+            new_schema[schema_key] = schema_registry.access(
+                new_schema[schema_key])
```
U8NWXD, line 508: What if `new_schema` is a dict but `new_schema[schema_key]` is not a string? Do we handle dict dividers and updaters elsewhere?

vivarium/core/store.py
U8NWXD, line 502: The naming here might be confusing because `current_schema` isn't really a schema. It's the value of one schema key

vivarium/core/store.py
```diff
@@ -476,24 +476,45 @@ def _check_default(self, new_default):
...
+        if (
+                current_schema
+                and current_schema != DEFAULT_SCHEMA
+                and current_schema != new_schema):
+            raise ValueError(
+                f"incompatible {schema_key} schema assignment: {new_schema} "
+                f"at {self.path_for()}. "
+                f"schema {current_schema} is already assigned.")
```
U8NWXD, line 516: I think this error message could be clearer, e.g.:

`Incompatible schema assignment at {self.path_for()}. Trying to assign the value {new_schema} to key {schema_key}, which already has the value {current_schema}.`

vivarium/core/store.py
```diff
@@ -476,24 +476,45 @@ def _check_default(self, new_default):
...
-        return new_value
+        current_schema = self.__dict__[schema_key]
+        if current_schema is not None and current_schema != new_schema:
+            raise ValueError(
+                f"incompatible {schema_key} schema assignment: {new_schema} "
+                f"at {self.path_for()}. "
+                f"schema {current_schema} is already assigned.")
+        return new_schema
+
+    def _check_schema_set_default(self, schema_key, new_schema, schema_registry):
```
U8NWXD, line 501: The `_set_default` part of the name confuses me since it doesn't look like we assign any defaults here

U8NWXD (changes requested):

vivarium/core/store.py
eagmon: how about `current_schema_value`?

vivarium/core/store.py
eagmon: maybe `_check_schema_methods`? This is for dividers/updaters, and can access these methods through their registries. This is different from the `_check_schema` method, which does not access methods through registries or check if the default schema value is being used.

U8NWXD: Maybe `_check_for_schema_conflicts_handle_defaults()` or `_check_schema_handle_defaults`? I don't love "handle" though since that still sounds like we are applying the defaults. Maybe "support"?

eagmon: `_check_schema_support_defaults` is good engough. I'll go with that.

vivarium/core/store.py
eagmon: I don't think we have supported dict dividers. We handle updater dicts here: https://github.com/vivarium-collective/vivarium-core/blob/48c6d0119d8b2192661289f25558ce33fdd39c18/vivarium/core/store.py#L695-L696

Issue #171, opened 2022-03-18
Improve Handling of Process Parameters and Defaults


U8NWXD:
## Current Approach

Currently, processes have a `defaults` class variable that specifies the parameters they expect. Then when calling a process, you can provide parameters in a dictionary to the `parameters` argument to override those defaults. For example:

```python
class MyProcess(Process):

    defaults = {
        'flag': True,
    }

proc = MyProcess({'flag': False})
```

## Problem

Our current approach for handling process parameters and defaults has a number of issues:

1. Parameters do not have types, which prevents type checking from catching typing errors. For example, you could write `MyProcess({'flag': 'test'})` without raising any type errors.
2. Similarly, we don't actually enforce the convention that all accepted parameters be present in `defaults`. For example, `MyProcess` could have a `next_update` that uses `self.parameters['flag2']` even though `MyProcess.defaults` does not contain `flag2`.
3. Other Python tools don't understand our way of specifying parameters. This means that e.g. PyCharm won't suggest parameters when you initialize a process.
4. Our approach is not the standard Python way of specifying parameters, which is confusing.
5. `defaults` are not handled correctly when subclassing processes. For example, consider the process `class MyProcess2(MyProcess)` with a constructor that calls `MyProcess.__init__(parameters)`. Inside `MyProcess.__init__`, `self.defaults` refers to `MyProcess2.defaults`, not `MyProcess.defaults` like you'd expect.

## Functionality to Preserve

1. Backwards compatibility. We don't want to have to rewrite all our libraries.
2. Ability to save off parameters for serialization.

eagmon:
Some questions about the alternative @U8NWXD:

1. Is the only required backwards compatibility to support a `self.defaults` dict and a `parameters` dict being passed to the Process superclass with `super().__init__(parameters)`? Anything else?
2. Is the only purpose of having the `Process` superclass access the subclass locals to save these parameters for serialization?

U8NWXD:
Regarding https://github.com/vivarium-collective/vivarium-core/issues/171#issuecomment-1081064410:

1. I think we also need to handle of special parameters like `time_step` and `_original_parameters`, and we need to be able to access parameters through `self.parameters`.
[...]

PR #177, opened 2022-04-07
Process keywords

eagmon:
This draft PR includes work towards improved handling of Process parameters as keywords (#171)

repository.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -159,7 +159,8 @@ class Process(metaclass=abc.ABCMeta):
                 '_emit': True,
                 '_updater': 'set'}))
 
-        self.parameters.setdefault('time_step', DEFAULT_TIME_STEP)
+        self._set_timestep()
+
         self.schema: Optional[dict] = None
 
     def initialize(self, parameters, base_parameters):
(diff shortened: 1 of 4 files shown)


vivarium/processes/growth_rate.py
```diff
@@ -22,11 +22,22 @@ class GrowthRate(Process):
...
+    def __init__(
+            self,
+            default_growth_rate=0.0005,
+            default_growth_noise=0.0,
+            variables=('mass',),
+            **base_parameters):
+
+        parameters = locals().copy()
+        base_parameters = parameters.pop('base_parameters')
+        self.initialize(parameters, base_parameters)
```
prismofeverything, line 40: I know I helped create this thing, but on reflection I feel the cost here for `Process` creators to have to add this four-line litany to every `__init__` method is too high. I appreciate what we're trying to solve here but the trade-off can't be to create cryptic things we demand of our users.... IMO

vivarium/core/process.py
```diff
@@ -275,10 +302,10 @@ def calculate_timestep(self, states: Optional[State]) -> Union[float, int]:
         """Return the next process time step
 
         A process subclass may override this method to implement
-        adaptive timesteps. By default it returns self.parameters['time_step'].
+        adaptive timesteps. By default it returns self.parameters['timestep'].
         """
         _ = states
-        return self.parameters['time_step']
+        return self.parameters['timestep']
```
prismofeverything, line 308: `timestep` change is good, maybe a separate PR?

vivarium/processes/growth_rate.py
eagmon: Maybe we can make this optional? If people want function keywords, they can follow this template. But the previous `__init__` approach with defaults and a parameters dict can work just as well.

PR #184, opened 2022-04-20
Prototype of logging updates

U8NWXD:
I've used this for debugging, but it probably needs some more work to be merged in

repository.


eagmon:
This is an interesting idea. Similar to @Robotato 's `logging_process` in terms of its functionality, except that his added a new `logging_port` to selected processes, and saves their full updates to a new store. Your solution is simpler in that you can just switch logging on via the store schema. Some added functionality that `logging_process` has was that you could see what process each update came from -- you might be able to do that here if you can get the update's source.

PR #185, opened 2022-04-22
timestep instead of time_step

eagmon:
This change the process parameter to `timestep` from `time_step`. This change is to keep consistent with other parts of the codebase, such as `engine.py`. `time_step` as a process input parameter remains compatible, but is internal changed to `timestep`.

repository.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -82,43 +82,44 @@ def _get_parameters(
 
 
 class Process(metaclass=abc.ABCMeta):
+    """Process parent class.
+
+      All :term:`process` classes must inherit from this class. Each
+      class can provide a ``defaults`` class variable to specify the
+      process defaults as a dictionary.
+
+      Note that subclasses should call the superclass init function
+      first. This allows the superclass to correctly save the initial
+      parameters before they are mutated by subclass constructor code.
+      We need access to the original parameters for serialization to
+      work properly.
+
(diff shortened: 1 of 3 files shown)


vivarium/core/process.py
```diff
@@ -155,9 +156,15 @@ class can provide a ``defaults`` class variable to specify the
...
 
-        self.parameters.setdefault('time_step', DEFAULT_TIME_STEP)
+        self._set_timestep()
+
         self.schema: Optional[dict] = None
 
+    def _set_timestep(self) -> None:
+        self.parameters.setdefault('timestep', DEFAULT_TIME_STEP)
+        if self.parameters.get('time_step'):
+            self.parameters['timestep'] = self.parameters['time_step']
```
U8NWXD, line 166: Should we delete `time_step` from the parameters?
* Pros: We can be confident there isn't any code left relying on reading `time_step`
* Cons: If any process is trying to read `self.parameters['time_step']`, they'll break. This is technically a breaking change

eagmon: @U8NWXD I'm inclined to leave it, despite it being kind of clunky. I am personally never certain if it is "time step" or "timestep", and am pretty sure people use both spelling. So this clunky solution at least would avoid that ambiguity.

U8NWXD (approved):

PR #186, opened 2022-04-26
Apply function to data from emitter

thalassemia:
Lately, Chris and I have been running into some issues with exorbitant memory usage when retrieving emitted data from MongoDB. In both of our cases, the available retrieval tools are simply too heavy-handed in retrieving anything and everything matching a given query. 

For my particular use case, I found that by adding a transformation function parameter to `data_from_database`, I could greatly reduce memory usage by aggregating each document as desired. As an example, say I have 500 documents with the following field hierarchy:
```
{
    'A': {
        1: {'ID': 1, 'data': ...},
        2: {'ID': 2, 'data': ...},
        ...
    },
    'B': {...}
}
```
If I wanted to get the number of items in `A`, at present, my only option is to first retrieve everything under `A` for all 500 documents, all of which gets stored in memory. By passing an aggregation function like `len()` at retrieval time, I can apply `len()` to the `A` field for each document as it is being retrieved, reducing the final memory footprint.

repository.


vivarium/core/emitter.py
```diff
@@ -425,6 +439,9 @@ def get_history_data_db(
     for document in cursor:
         assert document.get('assembly_id'), \
             "all database documents require an assembly_id"
+        if f and query:
+            for field in query:
+                document["data"] = apply_func(document["data"], field, f)
```
U8NWXD, line 444: I can imagine that in lots of cases, users might want to only apply the function to some of the fields they want to retrieve. Can we support that? Some ideas:

1. Instead of just providing a function `f`, provide a dictionary mapping paths to functions.
2. Pass the path `field` to the function `f` so the function can behave differently depending on what field it's operating on

U8NWXD (changes requested):

vivarium/core/emitter.py
eagmon: I added a `func_dict` arg that does this.

PR #192, opened 2022-05-18
remove composition.py

eagmon:
This removes vivarium-core dependence from `composition.py`, and adds a deprecation warning. This file holds a collection of helper functions for bringing processes together. In the last year `composite.merge` and `Engine` have been made much easier to use directly instead of relying on helper functions. Because of this `composition.py` is no longer helpful and in some cases made upkeep harder. 

`directories.py` now has a few standard directories -- `PROCESS_OUT_DIR`, `COMPOSITE_OUT_DIR`, etc.

This will be a breaking change and I expect many dependent libraries will have to update their code to remove the helper functions. Some tips for this:

Instead of `simulate_process`:
```
    my_composite = my_process.generate()
    sim = Engine(composite=my_composite)
    sim.update(total_time)
    timeseries = sim.emitter.get_timeseries()
```

instead of `simulate_composer`:
```
    sim = Engine(
        composite=composite,
        initial_state=initial_state
        )
    sim.update(time_total)
    output = sim.emitter.get_timeseries()
```

adding a timeseries process:
```
    from vivarium.processes.timeline import TimelineProcess
    # create the processes
    agent = ToyAgent({'agent_id': agent_id})
    timeline_process = TimelineProcess({'timeline': timeline})

    # compose
    composite = agent.generate(path=('agents', agent_id))
    composite.merge(
        processes={'timeline': timeline_process},
        topology={'timeline': {
            'global': ('global',),
            'agents': ('agents',)}},
    )
```

repository.


eagmon:
@U8NWXD -- This could go into a 2.0 release, but it alone would not warrant a 2.0 release. What is the best practice for working on a number of changes and then releasing them all at once as a new version? Maybe this should go into a dev branch?

U8NWXD:
I looked into this and proposed a workflow in https://github.com/vivarium-collective/vivarium-core/issues/194 for developing v2 while still maintaining v1

eagmon:
@U8NWXD -- I'm in support of your proposal for v2. But to put that off for a bit I brought `composition.py` back in this PR with a deprecation warning. So the main change for this PR is that Vivarium-core no longer uses `composition.py`, even though it is still available for dependent libraries. This will make it easier to remove when the time comes.

Issue #197, opened 2022-05-27
Proposal: Improved Handling of Parallel Processes


U8NWXD:
### Message Passing

When a process has `Process.parallel` set to true, we launch a parallel OS process that runs `vivarium.core.process._run_update()`. `run_update()` contains a loop that repeatedly receives `(interval, state)` tuples from a pipe, passes those arguments to `Process.next_update()`, and sends the result back through the pipe. To stop the parallel process, we pass `interval=-1`.

### Tracking in Main OS Process

In the main OS process (which contains `Engine`), we store `ParallelProcess` instances in `Engine.parallel`. Then when we need an update from a process, we call `Engine._invoke_process`, which does the following:

* If the process is parallel, starts the `ParallelProcess` computing the update and returns it.
* If the process is non-parallel, creates and returns an instance of `InvokeProcess`, which has an interface similar to `ParallelProcess` but computes the update immediately. Note that there's an extra `invoke_process` function that computes the process's update, but this extra level of indirection appears unnecessary.

## Problems

The way we currently track parallel processes has a number of downsides:

1. We store `ParallelProcess` instances in `Engine.parallel`, but we also store the original `Process` instances in the store hierarchy (with a reference in `self.processes`).
   1. This is unnecessary and wastes memory, especially for processes with large parameter dictionaries or internal states.
   2. Having the original `Process` instances in the store hierarchy is confusing. A user can read out the internal state of those processes with no problem, but they're getting the state from a process that hasn't changed since the beginning of the simulation, which is not intuitive.
2. There's no way to retrieve any information from a process besides its next update. This is a problem when we want to read out a process's internal state (e.g. for debugging or when working with inner simulations).
3. The extra levels of indirection are confusing. Every time I work on this, I have to trace through the code again to remind myself what all the "invoke" things do.

## Proposed Solution

1. Eliminate the extra `invoke_process` function that doesn't appear to do anything.
2. Instead of storing `ParallelProcess` instances in `self.parallel`, put them directly into the store hierarchy with references in `self.processes`. Once a parallel process has been put on its own OS process, there should be no copies of it left in the main OS process.
3. Systematize message-passing between the main and parallel OS processes as commands with arguments. `Process` will have a `run_command()` method that accepts a command name (string) and a tuple of arguments. The `_run_update()` function would handle the following commands:

   1. `_halt`: Ignores arguments and shuts down the parallel process (equivalent of passing `interval=-1` currently)
[...]

PR #198, opened 2022-05-27
Process Commands

U8NWXD:
Introduce process commands to support more interactions with parallel processes. Now all `Process` methods of a parallelized process can be queried from the parent OS process. Users can also add support for custom methods of their processes.

This change also simplifies the way `Engine` handles parallel processes warns users when serializers are not being found efficiently.

TODO:

* [x] Add safety checks to make sure users don't send commands or get command results out of order
* [x] Make sending commands private?
   * Decided not to make this private because since we want users to be able to use commands for subclass methods, commands need to be part of the public API
* [x] Thoroughly test parallelization and process commands
* [x] Change `Engine` to put `ParallelProcess` instances in the store hierarchy
* [x] Eliminate `invoke_process` function that doesn't do anything

repository.


vivarium/core/process.py
```diff
@@ -544,9 +604,10 @@ def update(
                 should be computed.
             states: The pre-update state of the simulation.
         """
-        self.parent.send((interval, states))
+        args = (interval, states)
+        self.send_command('_next_update', args)
 
-    def get(self) -> Update:
```
eagmon, line 549: no more `.get()`? Does this break an prior work?

U8NWXD: This is in `ParallelProcess`, so I think this is probably only used inside Vivarium Core. I doubt any clients are using this, but I'm not sure. That said, I haven't checked to make sure I've updated all the Vivarium Core usages yet.

Also note that this is a draft PR because I think we should discuss on #197 whether we even want to take this approach.

vivarium/core/process.py
```diff
@@ -143,29 +151,157 @@ def __init__(self, parameters: Optional[dict] = None) -> None:
         elif not hasattr(self, 'name'):
             self.name = self.__class__.__name__
 
-        self.parameters = copy.deepcopy(self.defaults)
-        self.parameters = deep_merge(self.parameters, parameters)
-        self.schema_override = self.parameters.pop('_schema', {})
-        self.parallel = self.parameters.pop('_parallel', False)
-        self.condition_path = None
+        self._parameters = copy.deepcopy(self.defaults)
+        self._parameters = deep_merge(self._parameters, parameters)
```
eagmon, line 155: we use `self.parameters` all over, not sure this `_` to make it private will be backwards-compatible.

vivarium/core/engine.py
```diff
@@ -711,7 +711,8 @@ def _invoke_process(
                 self.parallel[path] = ParallelProcess(
                     process, bool(self.profiler))
             # trigger the computation of the parallel process
-            self.parallel[path].update(interval, states)
+            self.parallel[path].send_command(
+                'next_update', (interval, states))
```
eagmon, line 715: this is very elegant! `send_command()` will greatly expand how we use parallel processes. great.

vivarium/core/process.py
```diff
@@ -143,29 +151,157 @@ def __init__(self, parameters: Optional[dict] = None) -> None:
...
+    def parallel(self) -> bool:
+        return self._parallel
+
+    @property
+    def condition_path(self) -> Optional[HierarchyPath]:
+        return self._condition_path
+
+    @property
+    def schema(self) -> Optional[Schema]:
+        return self._schema
```
eagmon, line 192: just to make sure I understand -- these `@property` will allow parallel processes to be inspected by calling these methods rather than directly inspecting the variables (which are now private)

eagmon: can you call `properties` without the parentheses? `self.parameters()` would be a pretty big breaking change.

U8NWXD: Reading these properties works exactly like before. If you have a process `p`, then `p.schema` will evaluate to the result of calling the `schema()` method of `p`. (I'm not sure `p.schema()` will even work). That should mean there's no breaking change

eagmon (approved):
This is looking really good! I have no improvements. `Engine` is actually simpler now in many ways. `Process` is a bit more complicated, but I think it all makes sense and is required.

@prismofeverything -- Can you give this a look, since it is a major design decision?

doc/guides/processes.rst
```diff
@@ -562,3 +562,40 @@ The above pseudocode is simplified, and for all but the most simple
...
+
+In normal situations though, you shouldn't have to worry about process
+commands. Instead, just pass ``'_parallel': True`` in a process's
+configuration dictionary, and the Vivarium Engine will handle the
+parallelization for you. Just remember that parallelization requires
+that processes be serialized and deserialized at the start of the
+simulation, and this serialization only preserves the process
+parameters. This means that if you instantiate a process and then change
+its instance variables, those changes won't be preserved when the
+process gets parallelized.
```
prismofeverything, line 601: Excellent disclaimer, glad this is documented.

vivarium/core/engine.py
```diff
@@ -119,21 +120,6 @@ def timestamp(dt: Optional[Any] = None) -> str:
...
-    """Compute a process's next update.
-
-    Call the process's
-    :py:meth:`vivarium.core.process.Process.next_update` function with
-    ``interval`` and ``states``.
-    """
-
-    return process.next_update(interval, states)
-
-
```
prismofeverything, line 136: Ha! Who knows what this ever did.

vivarium/core/engine.py
```diff
@@ -188,44 +176,6 @@ def get(self) -> Update:
         return {}
 
 
-class InvokeProcess:
-    def __init__(
```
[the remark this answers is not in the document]
prismofeverything: Great simplification.... sometimes indirection is a form of superstition perhaps.

vivarium/core/engine.py
```diff
@@ -694,28 +667,16 @@ def _invoke_process(
...
-            if path not in self.parallel:
-                self.parallel[path] = ParallelProcess(
-                    process, bool(self.profiler))
-            # trigger the computation of the parallel process
-            self.parallel[path].update(interval, states)
-
-            return self.parallel[path]
-        # if not parallel, perform a normal invocation
-        return self.invoke(process, interval, states)
+        process.send_command('next_update', (interval, states))
```
prismofeverything, line 678: This is pretty nice, `send_command` becomes transparent to the `Engine` this way.

vivarium/core/process.py
```diff
@@ -120,6 +122,14 @@ class can provide a ``defaults`` class variable to specify the
       """
 
     defaults: Dict[str, Any] = {}
+    METHOD_COMMANDS = (
+        'initial_state', 'generate_processes', 'generate_steps',
+        'generate_topology', 'generate_flow', 'merge_overrides',
+        'calculate_timestep', 'is_step', 'get_private_state',
+        'ports_schema', 'next_update', 'update_condition')
```
prismofeverything, line 129: Quality list. I don't see anything missing (!) didn't realize the `Process` API had grown so much.

vivarium/core/process.py
prismofeverything: Interesting, I haven't used these before. I try to avoid special features like this, but this is a good use case really.

vivarium/core/process.py
```diff
@@ -143,29 +153,169 @@ def __init__(self, parameters: Optional[dict] = None) -> None:
...
+
+    def get_command_result(self) -> Any:
+        '''Retrieve the result from the last-run command.
+
+        Returns:
+            The result of the last command run. Note that this method
+            should only be called once immediately after each call to
+            :py:meth:`send_command`, and there should be no intervening
+            calls to :py:meth:`send_command`. When this order is
+            violated, behavior is undefined.
```
prismofeverything, line 301: Mildly concerning.... is there a way to make it so the user can't make this mistake?

prismofeverything: Or perhaps this method should be private.

vivarium/core/process.py
```diff
@@ -485,24 +635,43 @@ def is_step(self) -> bool:
...
 
-        # stop process by sending -1 as the interval
-        if interval == -1:
+        if command == 'end':
             running = False
-
         else:
-            update = process.next_update(interval, states)
-            connection.send(update)
+            result = process.run_command(command, args, kwargs)
```
prismofeverything, line 673: Excellent

vivarium/core/process.py
```diff
@@ -619,3 +906,44 @@ def test_serialize_process_inheritance() -> None:
...
+    parallel_proc.send_command('compare_pid', (os.getpid(),))
+    assert not parallel_proc.get_command_result()
+
+
+def test_invalid_command() -> None:
+    proc = ToyParallelProcess()
+    with pytest.raises(ValueError) as exception:
+        proc.send_command('missing_command')
+    msg = 'does not understand the process command missing_command'
+    assert msg in str(exception.value)
```
prismofeverything, line 949: Appreciate the extensive tests.

vivarium/library/dict_utils.py
```diff
@@ -314,6 +314,22 @@ def make_path_dict(embedded_dict):
     return path_dict
 
 
+def apply_func_to_leaves(root: Any, func: Callable[[Any], None]) -> None:
```
prismofeverything, line 317: Love a good dictionary operation, and parametrizing functions (!) Solid addition to `dict_utils` (which could maybe be its own library if it grows large enough.... with a better name).

prismofeverything (approved):
Hey Chris, this is an excellent feature! I think we should keep it, and treat our processes as "services" essentially in this way, which sets the groundwork for a fully distributed implementation. Some minor comments but really it's good to go.

vivarium/core/engine.py
```diff
@@ -592,6 +550,22 @@ def _make_store(
             self.flow = self.state.get_flow() or {}
             self.topology = self.state.get_topology()
 
+    def _parallelize_processes(
+            self, processes: Any) -> Union[dict, Process]:
```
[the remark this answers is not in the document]
U8NWXD: Just to document my thought process, I decided to swap processes for `ParallelProcess` instances in `Engine` instead of in `Store` because we need to know whether to enable the profiler when creating the `ParallelProcess`. `Engine` has that information, but it felt wrong to pass that information through to all the `Store` nodes.

vivarium/core/process.py
U8NWXD: Fixed in https://github.com/vivarium-collective/vivarium-core/pull/198/commits/b1bd834d6391c9eff503421b8223cfc30f8235f8. Now an error will be raised if users try to run these commands out of order

PR #201, opened 2022-06-20
Add initial_state to composition.py Engine keys

U8NWXD:
In c5ce1b4, `initial_state` was removed from the list of allowed
experiment config keys to pass to `Engine`. This change was probably
made to avoid a clash between `initial_state` in the config dictionary
and the keyword argument, but the consequence is that initial states can
no longer be passed to many `composition.py` functions.

This commit restores `initial_state` to the list of allowed config keys
and avoids the clash by putting the keyword-arg `initial_state` into the
config dictionary.

repository.


eagmon (approved):
lgtm!

PR #204, opened 2022-07-27
Optimize Serialization Using BSON C Extensions

thalassemia:
## Rationale
Serialization of emitter data can make up a large portion of simulation runtimes, according to profiling data. 
## Proposal
PyMongo implements its serialization logic in C and has the ability to accept custom encoders via the [`TypeCodec` interface](https://pymongo.readthedocs.io/en/stable/examples/custom_type.html). Here, I modify the `Serializer` class to instantiate and return `TypeCodec` instances for serialization. 
## Results
When no document splitting is required, this PR nearly halves simulation time (running a simulation with a relatively large and complex emit). When document splitting is required at every timestep, the performance improvement drops to about 20%.
## Drawbacks
- Our [document-splitting algorithm](https://github.com/vivarium-collective/vivarium-core/blob/02b6bb51717e9d760cf425f851ff1b9064eb7590/vivarium/core/emitter.py#L50) is very slow on unserialized data due to the costliness of converting Numpy arrays to strings (used to determine size of dictionary). In my simulations, splitting the emit before serialization is still marginally faster than what we currently have. However, I can see this resulting in performance regressions for simulations that emit more/larger Numpy arrays.
  - **Solution 1:** Use PyMongo's C extensions to convert the data as follows before splitting: `dict` --> `BSON` --> `dict`. The final dictionary will contain only BSON-supported types (e.g. built-ins), making the cost of our document splitting algorithm negligible. Interestingly, this round-trip scheme still yields a 40% improvement.
  - **Solution 2:** Try to serialize and insert each document without splitting. Default to the round-trip solution when splitting is necessary. This is implemented in 973de3a and yields the aforementioned 50% improvement. If every document requires splitting (worst case scenario), this is just Solution 1 with the added cost of serializing the entire emit at each timestep.
  - **Solution 3:** Serialize the emit to BSON, then check whether the BSON is over MongoDB's size limit. If so, deserialize it back to dict, split as needed, and send the split data to PyMongo for re-serialization and insertion. If not, instantiate a [`RawBSONDocument`](https://pymongo.readthedocs.io/en/stable/api/bson/raw_bson.html) instance (which avoids PyMongo's internal serialization logic) and insert. This was implemented in b6b46f6 and ended up being significantly slower than Solution 2 even in the case where documents are split at every timestep.
  - **Note:** All relative performance numbers come from running my particular simulation. These margins of improvement may increase, decrease, or even become negative when running other simulations.
- PyMongo's serializer requires that all dictionary keys be strings (or subclasses of strings, like `np.str_`). This is in line with the official JSON specifications, but it does break [this test](https://github.com/vivarium-collective/vivarium-core/blob/02b6bb51717e9d760cf425f851ff1b9064eb7590/vivarium/core/serialize_test.py#L125).
- PyMongo does not automatically serialize objects that are subclasses of a class defined in a `TypeCodec`. This means that each `Process`, `Composer`, etc. requires its own `TypeCodec` as [demonstrated here](https://github.com/vivarium-collective/vivarium-core/blob/02b6bb51717e9d760cf425f851ff1b9064eb7590/vivarium/experiments/engine_tests.py#L1138).
- `TypeCodec` does not support custom de/serializers for Python's built-in types.
- All Python objects of a given type must be serialized the same way (e.g. cannot treat float-filled and string-filled Numpy arrays differently).
- Performance can be bottlenecked by document splitting for very large emits (e.g. simulation after multiple generations).

repository.

### DIFF
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -11,7 +11,6 @@ import uuid
 from typing import Any, Dict, List, Optional, Tuple, Callable
 from urllib.parse import quote_plus
 
-from bson import _dict_to_bson, _bson_to_dict
 from pymongo.errors import DocumentTooLarge
 from pymongo import MongoClient
 
@@ -28,7 +27,10 @@ from vivarium.library.topology import (
     without_multi,
 )
 from vivarium.core.registry import emitter_registry
-from vivarium.core.serialize import get_codec_options, deserialize_value
+from vivarium.core.serialize import (
+    get_codec_options,
+    serialize_value,
+    deserialize_value)
 
 MONGO_DOCUMENT_LIMIT = 1e7
 
@@ -265,8 +267,7 @@ class RAMEmitter(Emitter):
                 if key not in ['time']}
             data_at_time = assoc_path({}, self.embed_path, data_at_time)
             self.saved_data.setdefault(time, {})
-            data_at_time = _dict_to_bson(data_at_time, False, self.codec_options)
-            data_at_time = _bson_to_dict(data_at_time, self.codec_options)
+            data_at_time = serialize_value(data_at_time, self.codec_options)
             deep_merge_check(
                 self.saved_data[time], data_at_time, check_equality=True)
 
@@ -375,9 +376,8 @@ class DatabaseEmitter(Emitter):
         # getting string representation of Numpy arrays
         except DocumentTooLarge:
             emit_data.pop('assembly_id')
-            data = _dict_to_bson(emit_data, False, self.codec_options)
-            data = _bson_to_dict(data, self.codec_options)
-            broken_down_data = breakdown_data(self.emit_limit, data)
+            emit_data = serialize_value(emit_data, self.codec_options)
+            broken_down_data = breakdown_data(self.emit_limit, emit_data)
             for (path, datum) in broken_down_data:
                 d: Dict[str, Any] = {}
                 assoc_path(d, path, datum)
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -298,6 +298,7 @@ class SharedRamEmitter(RAMEmitter):
         # We intentionally don't call the superclass constructor because
         # we don't want to create a per-instance ``saved_data``
         # attribute.
+        self.codec_options = get_codec_options()
         self.embed_path = config.get('embed_path', tuple())
 
 
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -347,7 +347,7 @@ class DatabaseEmitter(Emitter):
             data = _bson_to_dict(data, self.codec_options)
             broken_down_data = breakdown_data(self.emit_limit, data)
             for (path, datum) in broken_down_data:
-                d = {}
+                d: Dict[str, Any] = {}
                 assoc_path(d, path, datum)
                 d['assembly_id'] = assembly_id
                 table.insert_one(d)
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -12,8 +12,7 @@ from typing import Any, Dict, List, Optional, Tuple, Callable
 from urllib.parse import quote_plus
 
 from bson import _dict_to_bson, _bson_to_dict
-from bson.errors import BSONError
-from bson.raw_bson import RawBSONDocument
+from pymongo.errors import DocumentTooLarge
 from pymongo import MongoClient
 
 from vivarium.library.units import remove_units
@@ -336,20 +335,22 @@ class DatabaseEmitter(Emitter):
         Break up large emits into smaller pieces and emit them individually
         """
         assembly_id = str(uuid.uuid4())
-        emit_data['assembly_id'] = assembly_id
-        data = _dict_to_bson(emit_data, False, self.codec_options)
-        if len(data) > self.emit_limit:
+        try:
+            emit_data['assembly_id'] = assembly_id
+            table.insert_one(emit_data)
+        # If document is too large, serialize and deserialize using
+        # BSON C extension before splitting to avoid heavy cost of
+        # getting string representation of Numpy arrays
+        except DocumentTooLarge:
+            emit_data.pop('assembly_id')
+            data = _dict_to_bson(emit_data, False, self.codec_options)
             data = _bson_to_dict(data, self.codec_options)
-            data.pop('assembly_id')
             broken_down_data = breakdown_data(self.emit_limit, data)
             for (path, datum) in broken_down_data:
                 d = {}
                 assoc_path(d, path, datum)
                 d['assembly_id'] = assembly_id
                 table.insert_one(d)
-        else:
-            d = RawBSONDocument(data)
-            table.insert_one(d)
 
     def get_data(self, query: list = None) -> dict:
         return get_history_data_db(self.history, self.experiment_id, query)
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -13,6 +13,7 @@ from urllib.parse import quote_plus
 
 from bson import _dict_to_bson, _bson_to_dict
 from bson.errors import BSONError
+from bson.raw_bson import RawBSONDocument
 from pymongo import MongoClient
 
 from vivarium.library.units import remove_units
@@ -335,23 +336,20 @@ class DatabaseEmitter(Emitter):
         Break up large emits into smaller pieces and emit them individually
         """
         assembly_id = str(uuid.uuid4())
-        try:
-            emit_data['assembly_id'] = assembly_id
-            table.insert_one(emit_data)
-        # If document is too large, serialize and deserialize using
-        # BSON C extension before splitting to avoid heavy cost of
-        # getting string representation of Numpy arrays
-        except BSONError:
-            emit_data.pop('assembly_id')
-            data = emit_data.pop('data')
-            data = _dict_to_bson(data, False, self.codec_options)
+        emit_data['assembly_id'] = assembly_id
+        data = _dict_to_bson(emit_data, False, self.codec_options)
+        if len(data) > self.emit_limit:
             data = _bson_to_dict(data, self.codec_options)
+            data.pop('assembly_id')
             broken_down_data = breakdown_data(self.emit_limit, data)
             for (path, datum) in broken_down_data:
-                d = dict(emit_data)
-                assoc_path(d, ('data',) + path, datum)
+                d = {}
+                assoc_path(d, path, datum)
                 d['assembly_id'] = assembly_id
                 table.insert_one(d)
+        else:
+            d = RawBSONDocument(data)
+            table.insert_one(d)
 
     def get_data(self, query: list = None) -> dict:
         return get_history_data_db(self.history, self.experiment_id, query)
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -265,6 +265,8 @@ class RAMEmitter(Emitter):
                 if key not in ['time']}
             data_at_time = assoc_path({}, self.embed_path, data_at_time)
             self.saved_data.setdefault(time, {})
+            data_at_time = _dict_to_bson(data_at_time, False, self.codec_options)
+            data_at_time = _bson_to_dict(data_at_time, self.codec_options)
             deep_merge_check(
                 self.saved_data[time], data_at_time, check_equality=True)
 
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -11,6 +11,8 @@ import uuid
 from typing import Any, Dict, List, Optional, Tuple, Callable
 from urllib.parse import quote_plus
 
+from bson import _dict_to_bson, _bson_to_dict
+from bson.errors import BSONError
 from pymongo import MongoClient
 
 from vivarium.library.units import remove_units
@@ -20,7 +22,7 @@ from vivarium.library.dict_utils import (
     deep_merge,
 )
 from vivarium.library.topology import assoc_path, get_in, paths_to_dict
-from vivarium.core.serialize import deserialize_value
+from vivarium.core.serialize import get_codec_options, deserialize_value
 from vivarium.core.registry import emitter_registry
 
 MONGO_DOCUMENT_LIMIT = 1e7
@@ -241,6 +243,7 @@ class RAMEmitter(Emitter):
     def __init__(self, config: Dict[str, str]) -> None:
         super().__init__(config)
         self.saved_data: Dict[float, Dict[str, Any]] = {}
+        self.codec_options = get_codec_options()
 
     def emit(self, data: Dict[str, Any]) -> None:
         """
@@ -251,6 +254,8 @@ class RAMEmitter(Emitter):
         if data['table'] == 'history':
             emit_data = data['data']
             time = emit_data['time']
+            emit_data = _dict_to_bson(emit_data, False, self.codec_options)
+            emit_data = _bson_to_dict(emit_data, self.codec_options)
             self.saved_data[time] = {
                 key: value for key, value in emit_data.items()
                 if key not in ['time']}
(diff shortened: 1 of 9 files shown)


thalassemia:
Ultimately, I decided on **Solution 2** and updated the documentation to reflect the new API. The issues of `TypeCodec` not supporting custom serialization of built-ins or different serializers for homotypic objects can be sidestepped by manually defining serializers for individual stores using the `_serializer` key. These stores will, however, lose the performance advantage of using the BSON C extensions. The issue of slow document splitting is unavoidable, and, in any case, this solution is still almost always significantly faster than the current code. To summarize, the remaining drawbacks are:

1. All dictionary keys must be strings or subclasses thereof.
2. Type matching is exact, meaning unique subclasses require their own codecs.

U8NWXD:
> To summarize, the remaining drawbacks are:
> 
>     1. All dictionary keys must be strings or subclasses thereof.
> 
>     2. Type matching is exact, meaning unique subclasses require their own codecs.

I think these downsides are pretty minor compared to the performance benefits this change will bring. We were already converting dictionary keys to strings prior to emitting them, so imposing downside (1) makes the data more consistent between emits and the Stores. Violating downside (2) currently throws warnings because it's less efficient, so we're already discouraging that behavior. I also can't think of any situations where you'd need to have different serializers for instances of the same type. If you did somehow need this functionality, you could just create a serializer for the type and have that serializer handle the two kinds of objects of that type differently. So overall, I'm in favor of this approach

vivarium/core/registry.py
```diff
@@ -335,117 +353,129 @@ def divide_null(state):
...
     def deserialize(self, data):
         """Deserialize some data.
 
-        Subclasses relying on
-        :py:meth:`vivarium.core.registry.Serializer.serialize_to_string`
-        should not override this function and instead override
+        This function should typically only be overwritten if all codecs in
+        the subclass form one-to-one mappings between BSON and Python types.
+        Otherwise, leave this function alone and only override
         :py:meth:`vivarium.core.registry.Serializer.deserialize_from_string`.
```
U8NWXD, line 472: In this case, the function should just `pass`, right? It seems strange to ask users to override a function just to make it `pass`. Can we have an attribute that specifies whether this serializer is handled by Pymongo or not and then have this function check that attribute?

vivarium/core/serialize.py
```diff
@@ -155,24 +70,7 @@ def deserialize(self, data: Any) -> List[Any]:
...
-        float key will be serialized as a string, which will then be
-        deserialized to a string.
-        '''
-        serialized = {}
-        for key, value in data.items():
-            if not isinstance(key, str):
-                key = str(key)
-            serialized[key] = serialize_value(value)
-        return serialized
+        super().__init__()
```
U8NWXD, line 73: You can just omit this `__init__()` right?

vivarium/core/serialize.py
```diff
@@ -155,24 +70,7 @@ def deserialize(self, data: Any) -> List[Any]:
 class DictSerializer(Serializer):  # pylint: disable=abstract-method
```
U8NWXD, line 70: How does this work without any defined codecs? Is it because PyMongo handles dicts for us?

vivarium/core/serialize.py
```diff
@@ -1,53 +1,19 @@
...
-from vivarium.library.dict_utils import remove_multi_update
 from vivarium.library.units import units
+from vivarium.core.registry import serializer_registry, Serializer
 
-
-# Serialization and deserialization functions that handle arbitrary data
-# types.
-
-
-def serialize_value(value: Any) -> Any:
```
U8NWXD, line 22: I think the `serialize_value` interface is worth keeping to avoid breaking the API. It can just call the BSON serialization functions

vivarium/core/serialize.py
```diff
@@ -184,113 +82,28 @@ def deserialize(self, data: dict) -> dict:
...
 
     def __init__(self) -> None:
-        super().__init__(
-            name='units',
-            exclusive_types=(type(units.fg), type(1 * units.fg)))
+        super().__init__(name='units')
 
-    def can_serialize(self, data: Any) -> bool:
-        return isinstance(data, (Quantity, Unit))
+    class Codec(TypeEncoder):
```
U8NWXD, line 91: Can we call this `UnitCodec` or something similar to better communicate how it and `Codec2` (rename to `QuantityCodec`) are different?

vivarium/core/serialize.py
```diff
@@ -340,103 +153,99 @@ def deserialize(  # pylint: disable=arguments-differ
...
+            return value.tolist()
 
-    def can_deserialize(self, data: Any) -> bool:
-        return False
-
-    def deserialize(self, data: Any) -> None:
+    def deserialize_from_string(self, data: str) -> None:
         raise NotImplementedError(
-            'Processes cannot be deserialized.')
+            f'{self} cannot be deserialized.')
```
U8NWXD, line 172: This is going to print `NumpySerializer cannot be deserialized`, which is confusing because it's not a NumpySerializer being deserialized. Maybe also worth noting that this should never happen since we never serialize something in the format that would cause this deserializer to be called.

U8NWXD (changes requested):
I left some comments--mostly I think that the interface can be made clearer

vivarium/core/serialize.py
thalassemia: Removed in 9860d8a.

vivarium/core/serialize.py
thalassemia: Addressed in 5394975.

vivarium/core/serialize.py
thalassemia: Addressed in 5394975.

vivarium/core/serialize.py
thalassemia: Yup, PyMongo does not allow you to create custom serializers for built-in types. The dictionary, identity, and sequence serializers are really only used for deserialization. I've renamed them as such in 5394975.

vivarium/core/serialize.py
thalassemia: Went through and removed all of these in 5394975.

vivarium/core/registry.py
thalassemia: Cleaned up API significantly in 5394975.

vivarium/core/registry.py
```diff
@@ -335,117 +353,129 @@ def divide_null(state):
...
+    class Codec(TypeCodec):
+        """Example TypeCodec.
+        
+        All subclasses must have one or more codecs. Note that type checking
+        is exact, meaning subclasses of defined types require their own codecs.
         """
-        raise NotImplementedError(
-            f'{self} has not implemented serialize_to_string().')
-
+        python_type = type(Dummy()) #: Python type to serialize
```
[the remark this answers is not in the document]
thalassemia: Cleaned up API significantly in 5394975.

vivarium/core/serialize_test.py
```diff
@@ -120,7 +150,7 @@ def test_serialization_full() -> None:
             "!ProcessSerializer[{'timestep': 1.0, '_name': "
             "'SerializeProcess'}]"
         ),
-        '1': True,
+        # '1': True,
```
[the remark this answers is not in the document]
thalassemia: Removed in 5394975.

vivarium/core/registry.py
```diff
@@ -60,12 +60,35 @@
...
+Each :term:`serializer` is defined as a class that follows the API we
+describe below. Vivarium uses these serializers to convert emitted data
+into a BSON-compatible format for database storage. Serializer names are
+registered in :py:data:`serializer_registry`, which maps these names to
+serializer subclasses.
+
+Serializer API
+==============
+
+Each serializer SHOULD implement the following methods and attributes:
```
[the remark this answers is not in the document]
thalassemia: Cleaned up API significantly in 5394975.

vivarium/core/registry.py
```diff
@@ -60,12 +60,35 @@
...
+
+1. One or more codecs in of the type :py:class:`bson.codec_options.TypeCodec`
+   or one of its subclasses
```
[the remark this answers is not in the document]
thalassemia: Cleaned up API significantly in 5394975.

vivarium/core/registry.py
```diff
@@ -335,130 +363,71 @@ def divide_null(state):
...
-        Returns:
-            The serialized data.
+        self.codecs = self.get_codecs()
+        for codec in self.codecs:
+            if hasattr(codec, 'bson_type') and hasattr(codec, 'transform_bson'):
+                self.deserialize = lambda x: x
+                self.can_deserialize = lambda _: False
+    
+    def get_codecs(self):
+        """Get list of codecs in serializer.
```
eagmon, line 404: can you add a sentence here on what codec are used for.

eagmon (approved):
This is looking great. thank you @U8NWXD for reviewing and @thalassemia for cleaning this up a lot. I have just one comment on documentation of codecs, but otherwise this looks good to merge.

vivarium/core/registry.py
thalassemia: Addressed in 7e14b75.

vivarium/core/registry.py
```diff
@@ -335,130 +363,73 @@ def divide_null(state):
...
-        :py:meth:`vivarium.core.registry.Serializer.deserialize_from_string`.
-
-        Args:
-            data: Serialized data to deserialize.
-
-        Returns:
-            The deserialized data.
+        """Given a string of the form ``!...[data here]``,
+        where ``...`` is ``self.name``, this returns the 
+        string inside the square brackets.
```
U8NWXD, line 424: Is the idea that serializers that don't use PyMongo codecs should override this function with their own deserialization logic? If so, that should be explained in the docstring

vivarium/core/registry.py
```diff
@@ -335,130 +363,73 @@ def divide_null(state):
...
-        Args:
-            data: Data to serialize.
-
-        Returns:
-            The serialized data.
+        """Serialize data using correct codec.
+        
+        This is typically only called in the case that individual stores
+        are assigned custom serializers. For maximum performance, serialization
+        should be left to PyMongo instead of calling this function.
```
U8NWXD, line 415: ditto: explain if/when this function should be overridden by subclasses

vivarium/core/emitter.py
```diff
@@ -262,6 +265,8 @@ def emit(self, data: Dict[str, Any]) -> None:
                 if key not in ['time']}
             data_at_time = assoc_path({}, self.embed_path, data_at_time)
             self.saved_data.setdefault(time, {})
+            data_at_time = _dict_to_bson(data_at_time, False, self.codec_options)
+            data_at_time = _bson_to_dict(data_at_time, self.codec_options)
```
U8NWXD, line 269: Can we use the `serialize.py` API here? As written, doing so would unnecessarily call `get_codec_options()` each time, so maybe let `serialize()` take in a list of codecs as an optional argument? If the codecs aren't provided, it could run `get_codec_options()`. Alternatively just run `get_codec_options()` a lot if it's not too expensive

vivarium/core/emitter.py
```diff
@@ -357,14 +366,23 @@ def write_emit(self, table: Any, emit_data: Dict[str, Any]) -> None:
...
+        try:
+            emit_data['assembly_id'] = assembly_id
+            table.insert_one(emit_data)
+        # If document is too large, serialize and deserialize using
+        # BSON C extension before splitting to avoid heavy cost of
+        # getting string representation of Numpy arrays
+        except DocumentTooLarge:
+            emit_data.pop('assembly_id')
+            data = _dict_to_bson(emit_data, False, self.codec_options)
+            data = _bson_to_dict(data, self.codec_options)
```
U8NWXD, line 379: ditto: use `serialize.py` API here if possible

U8NWXD (changes requested):
Just a few more comments

vivarium/core/registry.py
thalassemia: As is, this method works seamlessly with the unit conversion logic in the if block beginning on line 1022 of `store.py`. 

This should only ever be overriden in the case described, when individual stores are assigned to the serializer subclass via the `_serializer` key. For performance reasons, the `_serializer` key should only be defined in the case that the data in that store is of a type that either can't be serialized by a custom codec (i.e. built-ins) or already has a defined codec but needs to be serialized differently from other instances of that type. In that case, users can override this `serialize` method to define whatever serialization routine they desire. 

Note that the values returned by this method later undergo another round of serialization under PyMongo's codec system, potentially causing unexpected behavior if `serialize` returns data of a type that is modified by another codec before insertion into the database. Best practice would be for `serialize` to always return data of a built-in type with predictable serialization behavior that cannot be modified by the end user. 

I'll update the docstring with these details.

vivarium/core/registry.py
thalassemia: That's right. This method should only be overridden if the user also overrode `serialize` to enable custom serialization on a store-by-store basis (as opposed to all data of a specific type being serialized the same way). 

As an aside, the `TypeCodec` system allows users to define the `bson_type` attribute and `transform_bson` method to deserialize all BSON types the same way. Letting PyMongo perform all serialization and deserialization would be fastest. Unfortunately, our serializers do not represent a one-to-one mapping between python and BSON types (e.g. lists and arrays both serialize to lists), which is why a lot of them override the `deserialize` method and use the slow `deserialize_value` API.

vivarium/core/emitter.py
thalassemia: Good call. Addressed in 740a7ab.

vivarium/core/serialize.py
```diff
@@ -1,62 +1,39 @@
...
 def deserialize_value(value: Any) -> Any:
     compatible_serializers = []
     for serializer_name in serializer_registry.list():
         serializer = serializer_registry.access(serializer_name)
         if serializer.can_deserialize(value):
             compatible_serializers.append(serializer)
     if not compatible_serializers:
-        raise ValueError(
-            f'No deserializer found for {value}')
+        # Most likely a built-in type with no custom serializer/deserializer
```
thalassemia, line 35: Unless the user makes an error and forgets to override the `deserialize` and `can_deserialize` methods when they override the `serialize` method, all values that enter this if block should simply be built-in Python types now that I've removed the `IdentitySerializer`. I removed that serializer because we can no longer assume a catch-all regex pattern to avoid using it on strings generated by custom serializers. I'll add in some kind of check to make sure all three serialization methods are overriden at the same time.

vivarium/core/store.py
```diff
@@ -651,9 +657,6 @@ def _apply_config(self, config, source=None):
                     self.units = self.units or self.default[0].units
                     self.serializer = (self.serializer or
                                        serializer_registry.access('units'))
-                elif isinstance(self.default, np.ndarray):
```
thalassemia, line 654: NumPy arrays are handled by the PyMongo codec system so we can avoid the slow Python code for store-specific serialization.

vivarium/core/store.py
```diff
@@ -1021,13 +1024,11 @@ def emit_data(self):
...
             if self.serializer:
                 if isinstance(self.value, list) and self.units:
-                    return self.serializer.serialize(
-                        [v.to(self.units) for v in self.value])
+                    return [self.serializer.serialize(v.to(self.units))
+                            for v in self.value]
                 if self.units:
                     return self.serializer.serialize(
                         self.value.to(self.units))
-                if isinstance(self.value, list):
```
thalassemia, line 1029: Matt added this in a PR some time back to fix vivarium save states. The save and load state test in `vivarium-ecoli` ran just fine now with these lines removed.

U8NWXD (approved):
LGTM! One last thing to check--now that we're using BSON and Pymongo for serialization, can we still save states correctly? It's possible that now calling `serialize()` on the state results in data that is BSON-compatible but not JSON-compatible, breaking Matt's state-saving code

U8NWXD:
Also the type checks are failing

vivarium/core/serialize.py
thalassemia: I'm honestly not sure how to ensure that the user does not mess this up. I made the documentation for the serializer API (at the top of `registry.py`) very clear about what needs to be overriden in all possible cases. Hopefully that's enough to avoid any unexpected behavior from forgetting to override the correct methods.

thalassemia:
I looked over the save state code, and we do not call any of the serialization routines in vivarium-core when creating the save state JSON file. Running from a saved state appeared to cause issues before, forcing Matt to add those lines in the `self.serializer` if block. It seems to work just fine now though.

PR #205, opened 2022-08-03
Prevent multi-updates in initial_state

thalassemia:
When a single process has multiple ports into a store holding a dictionary, running the `initial_state` method of the `Composite` yields a multi-update for the initial value of that store. This PR adds a flag to instead populate the store with the correct initial dictionary.

repository.

### DIFF
--- a/vivarium/library/topology.py
+++ b/vivarium/library/topology.py
@@ -149,7 +149,7 @@ def dict_to_paths(root, d):
         return [(root, d)]
 
 
-def inverse_topology(outer, update, topology, inverse=None, initial=False):
+def inverse_topology(outer, update, topology, inverse=None, multi_updates=True):
     '''
     Transform an update from the form its process produced into
     one aligned to the given topology.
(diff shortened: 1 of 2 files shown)


vivarium/library/topology.py
```diff
@@ -149,7 +149,7 @@ def dict_to_paths(root, d):
         return [(root, d)]
 
 
-def inverse_topology(outer, update, topology, inverse=None):
+def inverse_topology(outer, update, topology, inverse=None, initial=False):
```
U8NWXD, line 152: It feels like the `multi_update` logic should be the exception for when we're inverting updates, and the default should be the behavior for generating initial states (i.e. no `multi_update`s). That's technically a breaking change though. What do you think @eagmon?

U8NWXD (changes requested):

vivarium/library/topology.py
eagmon: Hmm, so you are suggesting `initial=True` as the default? Yeah, that could break some things. I think it is fine as-is.

eagmon (approved):

vivarium/library/topology.py
eagmon: Maybe change the flag to 'muli_updates=True' so that it is the flag, but still keep it true to avoid API change?

thalassemia: In 97911c5, I gave the flag a more intuitive name without changing the API.

U8NWXD (approved):

PR #206, opened 2022-08-09
Support multiple emits for the same timestep

U8NWXD:
This allows e.g. multiple parallel `Engine` instances to send emits to
the same destination, e.g. the same database. So long as they use the
same experiment IDs and have synchronized times, the emits will be
merged upon retrieval.

repository.

### DIFF
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -257,13 +257,13 @@ class RAMEmitter(Emitter):
         if data['table'] == 'history':
             emit_data = data['data']
             time = emit_data['time']
-            time_data = {
+            data_at_time = {
                 key: value for key, value in emit_data.items()
                 if key not in ['time']}
-            time_data = assoc_path({}, self.embed_path, time_data)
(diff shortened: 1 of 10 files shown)


U8NWXD:
In conversation with @eagmon, decided to remove `Engine.complete()` and instead expect users to set `force_complete=True` when running `Engine.run_for()` at the end of the simulation

vivarium/core/emitter.py
```diff
@@ -251,9 +257,13 @@ def emit(self, data: Dict[str, Any]) -> None:
         if data['table'] == 'history':
             emit_data = data['data']
             time = emit_data['time']
-            self.saved_data[time] = {
+            time_data = {
```
eagmon, line 260: this can easily be interpreted as "data about time". Can we call it `data_at_time`?

eagmon (approved):
Looks great @U8NWXD! Thanks for walking me through it earlier. Just one comment, otherwise good to merge

PR #211, opened 2022-09-02
Fix memory leak and parallel process division

thalassemia:
Parallel processes were not properly saving certain class attributes, including `_parallel`, causing them to be lost upon pickling during division.

Additionally, we were not properly ending parallel processes when deleting stores, causing a memory leak. 

Both of these issues are fixed in this PR.

repository.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -155,8 +155,8 @@ class Process(metaclass=abc.ABCMeta):
 
         self._parameters = copy.deepcopy(self.defaults)
         self._parameters = deep_merge(self._parameters, parameters)
-        self._schema_override: Schema = self._parameters.pop('_schema', {})
-        self._parallel = self._parameters.pop('_parallel', False)
+        self._schema_override: Schema = self._parameters.get('_schema', {})
+        self._parallel = self._parameters.get('_parallel', False)
         self._condition_path: Optional[HierarchyPath] = None
         self._command_result: Any = None
         self._pending_command: Optional[
@@ -164,7 +164,7 @@ class Process(metaclass=abc.ABCMeta):
 
         # set up the conditional state if a condition key is provided
         if '_condition' in self._parameters:
-            self._condition_path = self._parameters.pop('_condition')
+            self._condition_path = self._parameters.get('_condition')
         if self._condition_path:
             self.merge_overrides(assoc_path({}, self._condition_path, {
                 '_default': True,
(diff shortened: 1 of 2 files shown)


vivarium/core/store.py
```diff
@@ -1064,13 +1073,14 @@ def _delete_path(self, path):
         if not path:
             self.inner = {}
             self.value = None
-            return self
+            return None
```
eagmon, line 1076: is this change required because `self` is deleted?

vivarium/core/store.py
```diff
@@ -1064,13 +1073,14 @@ def _delete_path(self, path):
...
         target = self.get_path(path[:-1])
         remove = path[-1]
         if remove in target.inner:
-            lost = target.inner[remove]
+            # End any parallel processes to be deleted
+            self.recursive_end_process(target.inner[remove])
             del target.inner[remove]
-            return lost
+            return None
```
eagmon, line 1083: likewise here, is `lost` deleted? If it is not, then we should try to keep the prior API

vivarium/core/process.py
```diff
@@ -155,16 +155,16 @@ def __init__(self, parameters: Optional[dict] = None) -> None:
...
+        self._parallel = self._parameters.get('_parallel', False)
         self._condition_path: Optional[HierarchyPath] = None
         self._command_result: Any = None
         self._pending_command: Optional[
             Tuple[str, Optional[tuple], Optional[dict]]] = None
 
         # set up the conditional state if a condition key is provided
         if '_condition' in self._parameters:
-            self._condition_path = self._parameters.pop('_condition')
+            self._condition_path = self._parameters.get('_condition')
```
eagmon, line 167: why the change from `pop` to `get`?

eagmon (approved):
Just a few questions before you merge.

vivarium/core/store.py
thalassemia: Nah I just thought it wasn't really necessary. We don't make use of the return value anywhere and this is a private function.

vivarium/core/store.py
thalassemia: Same as above.

vivarium/core/store.py
eagmon: Can you bring it back? It might be part of the store api developed with @WilliamIX. Then we're good to merge

eagmon:
Fantastic @thalassemia!

PR #212, opened 2022-09-06
Make TimelineProcess a Step

thalassemia:
This corrects `TimelineProcess` to apply updates at the time(s) indicated in the timeline. Before, specifying an update at t=0, for example, only results in an updated simulation state after a timestep has elapsed. 

repository.

### DIFF
--- a/vivarium/core/composition.py
+++ b/vivarium/core/composition.py
@@ -344,7 +344,7 @@ def test_composer_in_experiment() -> None:
     output = simulate_experiment(experiment, settings)
 
     # check that timeline worked
-    assert output['aaa']['a1'][6] == 10
+    assert output['aaa']['a1'][5] == 10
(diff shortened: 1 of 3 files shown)


eagmon (commented):
I don't think `Timeline` can work as a Step. I see you are using the `self.parameters['timestep']`, but this parameters is pretty meaningless in a Step -- it just runs between the other processes, so their time steps are entirely responsible for how often `Timeline` runs. Not sure we can currently get that information into `Timeline.next_update` for it to run correctly.

thalassemia:
That's true. Leaving this as is also makes perfect sense if I just think of the timeline as a set of times when actions will occur. In that case, the state should only reflect the results of these actions at the beginning of the following time steps. 

We'll just have to keep this delay in mind when using `TimelineProcesses`.

PR #214, opened 2022-09-09
Optimize dictionary key removal

thalassemia:
The `without` and `without_multi` functions are surprisingly slow. Creating a shallow copy of the dictionary and popping the desired keys is functionally equivalent but much faster.

I also removed a `pyplot` import that no longer seems to be necessary.

repository.

### DIFF
--- a/vivarium/core/store.py
+++ b/vivarium/core/store.py
@@ -586,11 +586,9 @@ class Store:
         config.pop('_output', None)
 
         if '*' in config:
-            config = config.copy()
             self._apply_subschema_config(config.pop('*'))
 
         if '_subschema' in config:
-            config = config.copy()
             subschema = config.pop('_subschema')
             if source:
                 self.sources[source] = subschema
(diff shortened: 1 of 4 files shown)


vivarium/core/store.py
```diff
@@ -582,47 +582,51 @@ def _apply_config(self, config, source=None):
...
 
         # remove _output special key. This is used only by schema_topology.
-        config = without(config, '_output')
+        config = config.copy()
+        config.pop('_output', None)
 
         if '*' in config:
-            self._apply_subschema_config(config['*'])
-            config = without(config, '*')
+            config = config.copy()
```
eagmon, line 589: Do we need to copy `config` every time here? shouldn't the `config = config.copy()` be sufficient to make sure we are not mutating the original config?

eagmon (approved):
interesting find... So iterating over `d.items()` is the problem here? Just one question about all those `config.copy()` -- are these all really necessary?

vivarium/core/store.py
thalassemia: Great point! That should make it even faster than before.

PR #215, opened 2022-09-14
Serialization fixes

thalassemia:
BSON has a hidden maximum serialized size of 2GB (refer to [this](https://github.com/mongodb/mongo-python-driver/blob/179efda31200b495beab6c2e94f365f5713aadc4/bson/buffer.c#L106)). In this PR, we switch to using the `orjson` package, which has no size limits and is generally faster than BSON anyways. Orjson also has the benefit of natively supporting Numpy arrays and types. 

An important caveat is `np.str_`. Orjson will complain if dictionary keys are of the type `np.str_` or if the data to serialize contains arrays with `np.str_` values. The latter case is handled by a fallback [Numpy array serializer](https://github.com/vivarium-collective/vivarium-core/blob/a8749ce3fd202b9d79c9a82f810e617ee3e6746a/vivarium/core/serialize.py#L104), but users must manually ensure that all dictionary keys are Python strings and not Numpy strings.

Additionally, this PR makes some tweaks to multiprocessing:
- `ParallelProcesses` no longer keep a reference to the original `Process` instance after initializing a separate OS process. This reduces RAM usage.
- When parallelization is enabled, process schemas may be lost in transit. If this happens, the process schema is now retrieved anew when rebuilding the topology view.

repository.

### DIFF
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -759,6 +759,10 @@ class ParallelProcess(Process):
         self._ended = False
         self._pending_command: Optional[
             Tuple[str, Optional[tuple], Optional[dict]]] = None
+        
+        # Uncomment if using spawn or forkserver
+        del self.process
+        del self.child
 
     def send_command(
             self, command: str, args: Optional[tuple] = None,
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -759,8 +759,7 @@ class ParallelProcess(Process):
         self._ended = False
         self._pending_command: Optional[
             Tuple[str, Optional[tuple], Optional[dict]]] = None
-        
-        # Uncomment if using spawn or forkserver
+
         del self.process
         del self.child
 
--- a/vivarium/core/process.py
+++ b/vivarium/core/process.py
@@ -103,14 +103,6 @@ class Process(metaclass=abc.ABCMeta):
               also contain the following special keys:
 
               * ``name``: Saved to ``self.name``.
-              * ``_original_parameters``: Returned by
-                ``__getstate__()`` for serialization.
-              * ``_no_original_parameters``: If specified with a value
-                of ``True``, original parameters will not be copied
-                during initialization, and ``__getstate__()`` will
-                instead return ``self.parameters``. This puts the
-                responsibility on the user to not mutate process
-                parameters.
               * ``_schema``: Overrides the schema.
               * ``_parallel``: Indicates that the process should be
                 parallelized. ``self.parallel`` will be set to True.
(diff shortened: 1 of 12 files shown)


thalassemia:
Previously, pickling and unpickling of processes could get quite convoluted and messy. During pickling, we saved the `parameters` instance variable (e.g. `self.parameters`). To unpickle, we called the `__init__` method on this saved `parameters` dictionary. This requires users to be very conscientious of the values contained within `self.parameters`, can be slow (depending on process size and complexity of its `__init__`), and causes unexpected results (like the bug noted above where process schemas are sometimes lost).

In 6c9a660, I reverted our custom process serialization code and was able to achieve a substantial performance improvement while rectifying a memory leak. To date, I have still not been able to find the exact source of this leak while running using this custom serialization code but can confirm that removing said code fixes the issue entirely.

vivarium/core/process.py
```diff
@@ -103,14 +103,6 @@ class can provide a ``defaults`` class variable to specify the
...
 
               * ``name``: Saved to ``self.name``.
-              * ``_original_parameters``: Returned by
-                ``__getstate__()`` for serialization.
-              * ``_no_original_parameters``: If specified with a value
-                of ``True``, original parameters will not be copied
-                during initialization, and ``__getstate__()`` will
-                instead return ``self.parameters``. This puts the
-                responsibility on the user to not mutate process
-                parameters.
```
eagmon, line 113: `_original_parameters` no longer needed? What change made this possible?

vivarium/core/serialize.py
```diff
@@ -185,99 +181,76 @@ def deserialize(  # pylint: disable=arguments-differ
...
-def get_codec_options() -> CodecOptions:
-    """Returns a set of options used for serializing and
-    deserializing BSON by collecting codecs from all registered
-    serializers."""
-    codecs = []
-    for serializer in serializer_registry.registry.values():
-        codecs += serializer.get_codecs()
-    return CodecOptions(type_registry=TypeRegistry(codecs))
+
+def make_default() -> Callable:
```
eagmon, line 223: can you name this something more descriptive? `make_default` is really not clear. Maybe `make_serializer_function`?

vivarium/core/serialize.py
```diff
@@ -185,99 +181,76 @@ def deserialize(  # pylint: disable=arguments-differ
...
-    """Serializer for ``np.bool_`` objects."""
-    class Codec(TypeEncoder):
-        python_type = np.bool_
-        def transform_python(self, value: np.bool_) -> bool:
-            return bool(value)
-
-    def get_codecs(self) -> List:
-        return [self.Codec()]
-
-class NumpyInt64Serializer(Serializer):
```
eagmon, line 213: What happens to these different numpy types now?

eagmon (changes requested):
Just a few questions/requests:

vivarium/core/serialize.py
thalassemia: Orjson natively supports Numpy arrays and types, with the caveat of `np.str_` as noted in my initial comment.

vivarium/core/serialize.py
thalassemia: Good idea. I’ll make that change as soon as I can.

vivarium/core/process.py
thalassemia: This was part of the convoluted process serialization code I was referring to in my most recent comment. Now that we aren’t implementing our own `__getstate__` and `__setstate__` pickling methods, we no longer have to bother with saving the parameter dictionaries, original or otherwise.

vivarium/core/serialize.py
eagmon: ah yes. that's a nice bonus improvement!

thalassemia:
I've addressed the review comments and added a more informative error message to help users diagnose serialization issues.

eagmon (approved):
Improved performance, cleaner implementation, code reduction, good documentation -- looks good to me!

PR #217, opened 2022-09-26
Create Mongo index for simulation time

thalassemia:
Our default collection indices are out of date. This PR removes unused indices and implements the correct index for simulation time, which should improve performance for queries targeting specific time points or ranges.

repository.

### DIFF
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -34,15 +34,11 @@ from vivarium.core.serialize import (
 MONGO_DOCUMENT_LIMIT = 1e7
 
 HISTORY_INDEXES = [
-    'time',
-    'type',
-    'simulation_id',
+    'data.time',
     'experiment_id',
 ]
 
 CONFIGURATION_INDEXES = [
-    'type',
-    'simulation_id',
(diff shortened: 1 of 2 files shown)


vivarium/core/emitter.py
```diff
@@ -34,15 +34,11 @@
 MONGO_DOCUMENT_LIMIT = 1e7
 
 HISTORY_INDEXES = [
-    'time',
-    'type',
-    'simulation_id',
+    'data.time',
```
eagmon, line 37: Interesting that `time` is no longer an index. Did you test that `data.time` lets you retrieve the nested field?

eagmon (approved):
just one question before you merge.

vivarium/core/emitter.py
thalassemia: Just tested it on my local machine and it works great. Here is the [relevant documentation](https://www.mongodb.com/docs/manual/core/index-single/#create-an-index-on-an-embedded-field).

eagmon: LGTM

PR #218, opened 2022-09-27
Create new MongoClient per OS process

thalassemia:
The MongoDB documentation recommends that only a single `MongoClient` be created per application, taking advantage of connection pooling to reduce latency and the number of new connections ([refer to docs](https://www.mongodb.com/docs/manual/administration/connection-pool-overview/)). At the same time, PyMongo is not fork-safe and [strongly recommends](https://pymongo.readthedocs.io/en/stable/faq.html#using-pymongo-with-multiprocessing) that a new `MongoClient` be created for each child process. This PR strikes a balance between these guidelines by caching `MongoClient` instances by PID, only spawning a new instance if running under a new PID. 

repository.

### DIFF
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -13,7 +13,7 @@ from typing import Any, Dict, List, Optional, Tuple, Callable
 from urllib.parse import quote_plus
 
 from pymongo.errors import DocumentTooLarge
-from pymongo import MongoClient
+from pymongo.mongo_client import MongoClient
 
 from vivarium.library.units import remove_units
 from vivarium.library.dict_utils import (
--- a/vivarium/core/emitter.py
+++ b/vivarium/core/emitter.py
@@ -6,6 +6,7 @@ Emitters
 Emitters log configuration data and time-series data somewhere.
 """
 
+import os
 import json
 import uuid
 from typing import Any, Dict, List, Optional, Tuple, Callable
@@ -306,6 +307,7 @@ class DatabaseEmitter(Emitter):
     >>> # emitter = DatabaseEmitter(config)
     """
     default_host = 'localhost:27017'
+    client_dict: Dict[int, MongoClient] = {}
 
     @classmethod
     def create_indexes(cls, table: Any, columns: List[str]) -> None:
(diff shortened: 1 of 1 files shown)


vivarium/core/emitter.py
```diff
@@ -320,9 +322,12 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
 
-        # create object instance of mongo client to prevent forking issues
-        self.client = MongoClient(
-            config.get('host', self.default_host))
+        # create new MongoClient per OS process
+        curr_pid = os.getpid()
+        if curr_pid not in self.client_dict:
+            self.client_dict[curr_pid] = MongoClient(
+                config.get('host', self.default_host))
+        self.client = self.client_dict[curr_pid]
```
eagmon, line 330: what happens if the `client_dict` is not provided? Can we make the `client_dict` optional?

vivarium/core/emitter.py
```diff
@@ -306,6 +307,7 @@ class DatabaseEmitter(Emitter):
     >>> # emitter = DatabaseEmitter(config)
     """
     default_host = 'localhost:27017'
+    client_dict: Dict[int, MongoClient] = {}
```
eagmon, line 310: Should this be passed in? How is the user expected to use this?

eagmon (commented):
some questions below.

vivarium/core/emitter.py
thalassemia: The user is not expected to use this at all. Creating and managing `MongoClients` should occur seamlessly behind the scenes the same way we previously handled the singleton `MongoClient`. See [here](https://github.com/vivarium-collective/vivarium-core/blob/c5220e721fc96116dcd2fcb8bb94a86604f804b8/vivarium/core/emitter.py#L314).

thalassemia: I’ll update the code to tie `client_dict` to the class instead of the instance.

eagmon: got it! looks good.

eagmon (approved):
great improvement!

PR #221, opened 2022-10-20
Allow breakdown document to lack data key

thalassemia:
This PR fixes crashes that occur when breaking down a large document results in a document that lacks the `data` key altogether.

repository.


eagmon (approved):
LGTM!

PR #223, opened 2022-11-09
update pint imports for 0.20.1

eagmon:
pint version 0.20.1 changed the location of `Quantity` and `Unit`, which led to import errors for some vivarium users. This PR adds try/except statements where these are imported to make the imports come from the correct modules.

repository.


prismofeverything (approved):
Sometimes your units library breaks backwards compatibility.

PR #226, opened 2023-05-23
support python 3.11

eagmon:
This PR is working to add Python 3.11 support for vivarium-core, including updating the `setup.py` classifiers, the GitHub workflows, and some requirements versions.

repository.


eagmon:
Current status: the `wrapt` library is trying to import `formatargspec` from the `inspect` module, but this function was deprecated in Python 3.5 and has been removed from Python 3.11.

thalassemia:
After updating the dependencies, I think everything now works for Python 3.9+. We could retain support for Python 3.8 by using trial-and-error to downgrade the versions of certain dependencies (Numpy, Pint, and maybe others). However, according to [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), many big scientific computing packages are dropping 3.8 at this point, so I think it makes sense for us to do the same.

I also corrected a bug where parallelized Steps were not being properly terminated at the end of a simulation.

eagmon:
This is fantastic! Thanks @thalassemia :-)

thalassemia:
Is this ready to merge? It says that code owner review is required, and there are also all the Python 3.8 tests that I removed since we are dropping support for that.

eagmon:
@thalassemia I think this looks good to merge. I am also blocked from merging since I opened this PR, but I can bypass protections if you think that's safe. Or @prismofeverything @U8NWXD could weigh in and approve.

thalassemia:
@eagmon It looks safe to me. I'm upgrading `vivarium-ecoli` to Python 3.11 right now, so this would be nice to have ASAP.

vivarium/core/process.py
```diff
@@ -384,6 +385,7 @@ def generate_topology(self, config: Optional[dict] = None) -> Topology:
 
     def generate_flow(self, config: Optional[dict] = None) -> Flow:
         _ = config
+        _ = self # Silence pylint no-self-use
```
prismofeverything, line 388: Does this mean this should be a static method instead? Seems like an indicator that it's warning us we're not using `self`.

prismofeverything (approved):
This all looks good to me! Just curious about all the `_ = self` lines which seem to point to making those static methods or methods outside the class, but otherwise all seem like solid changes. Impressive how many changes were required for a non-major-version update

vivarium/core/process.py
thalassemia: `self` is useful because users might want to pull attributes from `self` in their own implementations of these methods. On the flip side, I refactored three methods in `engine.py` to be standalone functions because they should never be overridden by end users and do not reference `self`.

eagmon:
@thalassemia -- should we remove tests for 3.8? They are holding up the checks

thalassemia:
Yeah, we should. I don't know how to do it on my end (or if I have access).

eagmon:
hmm, looks like they are stuck on there even though our workflow has been updated. I tried pushing a new commit to see if they would be removed. Let's just merge this when 3.9-3.11 pass.

eagmon:
@thalassemia -- merged, and made a new release 1.5.7. Thank you so much for keep Vivarium up-to-date!

PR #227, opened 2023-08-01
Make Process.default_state recursive

thalassemia:
Though it is rarely used, the `Process.default_state` method previously broke when a process' ports schema had ports that were either not nested at all or nested more than one level deep. This PR fixes that by making `default_state` recursive and adds a test to ensure it works.

repository.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v1.5.6
```
eagmon, line 3: should be 1.5.8 now

eagmon (approved):

thalassemia:
Ready to merge. I'm pretty sure the required statuses thing is something that was configured on GitHub itself under Settings > Branches > Edit master > Require status checks to pass before merging. I can see it for the `vivarium-ecoli` repo where I am an admin but not for this one.

Issue #228, opened 2023-08-04
Schema validation fails with Numpy arrays


thalassemia:
The equality comparisons [here](https://github.com/vivarium-collective/vivarium-core/blob/7bef56beb6615219e21d6b8962dca29e83082b71/vivarium/core/store.py#L502C49-L502C49) and [here](https://github.com/vivarium-collective/vivarium-core/blob/7bef56beb6615219e21d6b8962dca29e83082b71/vivarium/core/store.py#L531) throw the following error when any values are Numpy arrays:

    ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

For example, this occurs when the new and current schemas are `{'_divider': {'divider': 'set_value', 'config': {'value': np.zeros(10)}}}`. The above Exception is raised due to the `np.zeros(10)` buried deep within the `_divider` schema. A custom recursive function is needed to check all schema values for embedded Numpy arrays and compare them separately using `np.array_equal`.

PR #233, opened 2023-09-09
Schema fix

thalassemia:
- Addresses some bugs in `emitter.py`
  - [`DatabaseEmitter.write_emit`](https://github.com/vivarium-collective/vivarium-core/blob/a927c7abf943d02c6d9a9008dacd37b15a597caf/vivarium/core/emitter.py#L381C1-L381C1): Drop the `ObjectID` added by `insert_one` before chunking large docs
  - [`get_history_data_db`](https://github.com/vivarium-collective/vivarium-core/blob/a927c7abf943d02c6d9a9008dacd37b15a597caf/vivarium/core/emitter.py#L626): `start_time` and `end_time` now work when `cpus=1`
  - [`data_from_database`](https://github.com/vivarium-collective/vivarium-core/blob/a927c7abf943d02c6d9a9008dacd37b15a597caf/vivarium/core/emitter.py#L695C10-L695C10): Clarify to accept a `Database` instance and not a `MongoClient`
- Addresses #228 in `store.py`
- Ensures all tests return `None` by default (otherwise will raise errors in a future version of `pytest`)

repository.

### DIFF
--- a/vivarium/core/composition.py
+++ b/vivarium/core/composition.py
@@ -381,7 +381,7 @@ def test_process_deletion() -> None:
     assert masses == expected_masses
 
 
-def test_composer() -> Dict:
+def test_composer(return_value: bool = False) -> Optional[Dict]:
     toy_compartment = ToyCompartment({})
     settings = {
         'total_time': 10,
@@ -394,7 +394,9 @@ def test_composer() -> Dict:
                 'GLC': 0,
                 'MASS': 3,
                 'DENSITY': 10}}}
-    return simulate_composer(toy_compartment, settings)
+    if return_value:
+        return simulate_composer(toy_compartment, settings)
+    return None
 
 
 if __name__ == '__main__':
--- a/vivarium/experiments/store_api.py
+++ b/vivarium/experiments/store_api.py
@@ -1,3 +1,4 @@
+from typing import Optional, cast
 import pytest
 
 from vivarium.composites.toys import Qo, ToyProcess, ToyComposer
@@ -17,7 +18,7 @@ def get_toy_store() -> Store:
     return store2
 
 
-def test_insert_process() -> Store:
+def test_insert_process(return_value: bool = False) -> Optional[Store]:
     """Test Store.insert by adding a new process
 
     Return:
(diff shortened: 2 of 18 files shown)


vivarium/core/composition.py
```diff
@@ -394,7 +394,9 @@ def test_composer() -> Dict:
                 'GLC': 0,
                 'MASS': 3,
                 'DENSITY': 10}}}
-    return simulate_composer(toy_compartment, settings)
+    if return_value:
+        return simulate_composer(toy_compartment, settings)
+    return None
```
eagmon, line 399: Why all of these changes with `return_value`?

vivarium/experiments/store_api.py
```diff
@@ -36,25 +37,27 @@ def test_insert_process() -> Store:
...
-    return store
+    if return_value:
+        return store
+    return None
 
 
 def test_rewire_ports() -> None:
     """connect a process' ports to different store"""
-    store = test_insert_process()
+    store = cast(Store, test_insert_process(return_value=True))
```
eagmon, line 47: what is this fixing?

eagmon (approved):
Looks good @thalassemia, just a couple of questions about some of the changes you made. Are all of these `if return_value` necessary? Will this be required of tests moving forward?

vivarium/experiments/store_api.py
thalassemia: The cast fixes a `mypy` error. `test_insert_process` now returns either a `Store` or `None`, and `mypy` is not smart enough to realize on its own that this particular return value is a `Store`. It complains that we try to index into this value (e.g. `store[index]`) later on.

vivarium/core/composition.py
thalassemia: `pytest` warned me that all tests should return `None`. It said non-compliant tests would return errors in future versions of `pytest`. We use our test functions in many places where return values are expected. I added the `return_value` option so that values are returned only in these places and not when the tests are run by `pytest`.

vivarium/experiments/store_api.py
eagmon: sounds good to me!

PR #234, opened 2023-09-13
Recursively compare schemas during validation

### LINKED ISSUE #228: Schema validation fails with Numpy arrays

thalassemia:
This PR properly closes #228 by allowing Numpy arrays to be buried deep within ports schemas ([example for `_divider`](https://github.com/vivarium-collective/vivarium-core/blob/ab6346d369d9224e2df6945db74c143007246300/vivarium/core/engine.py#L1341-L1346)). Additionally, this fixes a bug I introduced in #233 which caused things like 0 and `np.zeros(10)` to be considered equivalent in ports schemas when they should not be.

To ensure that everything works as intended, I added [`test_numpy_schema_validation`](https://github.com/vivarium-collective/vivarium-core/blob/a0945835681e63d58bdbfc9d0b666e44dce0f4a4/vivarium/core/engine.py#L1242) to `engine.py`.

repository.

### DIFF
--- a/vivarium/library/dict_utils.py
+++ b/vivarium/library/dict_utils.py
@@ -5,7 +5,9 @@ from functools import reduce
 import operator
 import traceback
 from typing import Optional, Any, Callable
+import warnings
 
+import numpy as np
 from vivarium.library.units import Quantity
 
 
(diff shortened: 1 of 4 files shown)


vivarium/library/dict_utils.py
```diff
@@ -21,6 +23,43 @@ def merge_dicts(dicts):
     return merge
 
 
+def deep_compare(dct_1, dct_2, path=tuple()):
```
eagmon, line 26: `deep_compare`! I like it.

eagmon (approved):
Thanks @thalassemia! This looks good.

PR #237, opened 2024-02-02
Add tests for MongoDB and fix bugs

thalassemia:
Adding MongoDB to the test environment proved very straightforward thanks to [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers). I repurposed an existing script to test large emits, query capabilities, and manual insertion of documents. Along the way, I fixed three issues:

- `time` key was not being added to broken-down emits when `time` was 0
- Deleting configuration in `delete_experiment_from_database` failed when using multiple CPU processes
- Documents manually inserted using `data_to_database` were not retrievable by `data_from_database` unless a specific structure was adhered to

According to `pytest`, coverage of `emitter.py` went from 44% to 90% with this PR. Most of the remaining untested code is in the MongoDB Atlas functions.

repository.

### DIFF
--- a/vivarium/experiments/hierarchy_composite_division.py
+++ b/vivarium/experiments/hierarchy_composite_division.py
@@ -87,7 +87,7 @@ def test_hierarchy_update() -> None:
         emit_step=100.0)
 
     # run the experiment long enough to divide
-    hierarchy_experiment1.update(2000)
+    hierarchy_experiment1.update(3000)
 
     # check that the agents updated after division
     for n in ['processes', 'steps', 'topology']:  # TODO: add flow
--- a/vivarium/experiments/hierarchy_composite_division.py
+++ b/vivarium/experiments/hierarchy_composite_division.py
@@ -71,7 +71,7 @@ def test_hierarchy_update() -> None:
                 'Protein': {'X': 50 * units.mg / units.mL}}}}
 
     # make a txtl composite, embedded under an 'agents' store
-    txtl_composer = AgentDivision({})
+    txtl_composer = AgentDivision({'growth': {'default_growth_noise': 0}})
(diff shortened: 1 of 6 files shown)


vivarium/experiments/hierarchy_composite_division.py
```diff
@@ -71,7 +71,7 @@ def test_hierarchy_update() -> None:
                 'Protein': {'X': 50 * units.mg / units.mL}}}}
 
     # make a txtl composite, embedded under an 'agents' store
-    txtl_composer = AgentDivision({})
+    txtl_composer = AgentDivision({'growth': {'default_growth_noise': 0}})
```
thalassemia, line 74: This test randomly failed on me once, so I'm hoping this change makes it more reproducible.

vivarium/experiments/large_experiment.py
```diff
@@ -43,23 +48,24 @@ def generate_processes(self, config):
...
+                'arr_size': self.config['arr_size']})
             for process_id in self.process_ids}
     def generate_topology(self, config):
         return {
             process_id: {'port': ('store', process_id,)}
             for process_id in self.process_ids}
 
 
-def run_large_initial_emit():
+def test_large_emits():
```
eagmon, line 60: glad to see this used as a test!

eagmon (approved):
looks great @thalassemia! 

Just a question about the pytest workflow -- how does this mongo image work? Does github host this?

thalassemia:
The image is hosted on Docker Hub: https://hub.docker.com/_/mongo. In this PR, I pinned the 6.0.13 version of the image for no particular reason. We could set the tag to `latest` to automatically pull the newest image.

